PR #10, opened 2023-09-20
changes for tls 1.3 with FIPS ready

JacobBarthelmeh:
ZD16581

### DIFF
--- a/src/wp_wolfprov.c
+++ b/src/wp_wolfprov.c
@@ -997,6 +997,10 @@ int wolfssl_provider_init(const OSSL_CORE_HANDLE* handle,
     int ok = 1;
     OSSL_FUNC_core_get_libctx_fn* c_get_libctx = NULL;
 
+#ifdef HAVE_FIPS
+    PRIVATE_KEY_UNLOCK();
+#endif
+
     for (; in->function_id != 0; in++) {
         switch (in->function_id) {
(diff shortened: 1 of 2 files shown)


SparkiDev (approved):

src/wp_wolfprov.c
```diff
@@ -997,6 +997,10 @@ int wolfssl_provider_init(const OSSL_CORE_HANDLE* handle,
     int ok = 1;
     OSSL_FUNC_core_get_libctx_fn* c_get_libctx = NULL;
 
+#ifdef HAVE_FIPS
+    PRIVATE_KEY_UNLOCK();
```
SparkiDev, line 1001: There are unlocks and locks around every private key operation.
Why is this needed and is there a lock call?

SparkiDev (changes requested):

src/wp_wolfprov.c
JacobBarthelmeh: It's needed for the HKDF extract call. It could be placed just around that function call.

SparkiDev: Please do.
I don't want global unlocks going on.
Thanks!

JacobBarthelmeh: updated, thanks for the review

SparkiDev (approved):

PR #21, opened 2024-02-27
Fix compilation problem with FIPS code

bandi13:


SparkiDev (approved):

douzzer (changes requested):
I believe the test we want for these is
```
#if (!defined(HAVE_FIPS) || FIPS_VERSION_GE(5,3)) && (LIBWOLFSSL_VERSION_HEX >= 0x05006002)
```
i.e. we need to support the new way of doing things for FIPS >= 5.3.

douzzer (approved):

PR #33, opened 2024-06-26
Fix rsa sign verify

bandi13:
This is a proposed fix to ZD#18157.

### DIFF
--- a/src/wp_rsa_sig.c
+++ b/src/wp_rsa_sig.c
@@ -436,14 +436,14 @@ static int wp_rsa_signverify_init(wp_RsaSigCtx* ctx, wp_Rsa* rsa,
     if (ok) {
         ctx->saltLen = RSA_PSS_SALTLEN_AUTO;
 
-        if (wp_rsa_get_type(rsa) == RSA_FLAG_TYPE_RSA) {
+        if (wp_rsa_get_type(ctx->rsa) == RSA_FLAG_TYPE_RSA) {
             ctx->padMode = RSA_PKCS1_PADDING;
         }
         else {
             char* mdName;
             char* mgfMdName;
             ctx->padMode = RSA_PKCS1_PSS_PADDING;
-            wp_rsa_get_pss_mds(rsa, &mdName, &mgfMdName);
+            wp_rsa_get_pss_mds(ctx->rsa, &mdName, &mgfMdName);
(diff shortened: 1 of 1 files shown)


kareem-wolfssl (dismissed):

src/wp_rsa_sig.c
```diff
@@ -414,15 +414,19 @@ static int wp_rsa_signverify_init(wp_RsaSigCtx* ctx, wp_Rsa* rsa,
 {
     int ok = 1;
 
-    if (ctx->rsa != rsa) {
+    if ((ctx == NULL) || (rsa == NULL)) {
+        ok = 0;
+    }
+    if ((ok) && (ctx->rsa != rsa)) {
```
SparkiDev, line 420: No need for brackets around ok.

src/wp_rsa_sig.c
```diff
@@ -414,15 +414,19 @@ static int wp_rsa_signverify_init(wp_RsaSigCtx* ctx, wp_Rsa* rsa,
...
         wp_rsa_free(ctx->rsa);
         ctx->rsa = NULL;
         if (!wp_rsa_up_ref(rsa)) {
             ok = 0;
+        } else {
```
SparkiDev, line 425: Formatting:
```
}
else {
```

src/wp_rsa_sig.c
```diff
@@ -432,14 +436,14 @@ static int wp_rsa_signverify_init(wp_RsaSigCtx* ctx, wp_Rsa* rsa,
     if (ok) {
         ctx->saltLen = RSA_PSS_SALTLEN_AUTO;
 
-        if (wp_rsa_get_type(rsa) == RSA_FLAG_TYPE_RSA) {
+        if (wp_rsa_get_type(ctx->rsa) == RSA_FLAG_TYPE_RSA) {
```
SparkiDev, line 439: No need for this. ctx->rsa == rsa at this point and on.

SparkiDev (changes requested):

src/wp_rsa_sig.c
bandi13: As I mentioned in the commit message, this is just a defensive coding. Currently `ctx->rsa == rsa`, but there may be a future change above that inadvertently violates this. Additionally, this doesn't use too much more code space.

SparkiDev (approved):

PR #36, opened 2024-07-11
Release fixes

bandi13:
This is to get `make distcheck` to work successfully before our first release.

### DIFF
--- a/test/include.am
+++ b/test/include.am
@@ -9,6 +9,8 @@ check_PROGRAMS += test/unit.test
 noinst_PROGRAMS += test/unit.test
 DISTCLEANFILES += test/.libs/unit.test
 
+# Override the 'certs' directory with the toplevel source directory for 'make distcheck' to work
+test_unit_test_CPPFLAGS  = -DCERTS_DIR='"$(abs_top_srcdir)/certs/"'
 test_unit_test_SOURCES = \
 	test/test_aestag.c \
 	test/test_cipher.c \
--- a/test/include.am
+++ b/test/include.am
@@ -10,7 +10,7 @@ noinst_PROGRAMS += test/unit.test
 DISTCLEANFILES += test/.libs/unit.test
 
 # Override the 'certs' directory with the toplevel source directory for 'make distcheck' to work
-test_unit_test_CPPFLAGS  = -DCERTS_DIR='"$(abs_top_srcdir)/certs/"'
+test_unit_test_CPPFLAGS  = -DCERTS_DIR='"$(abs_top_srcdir)/certs"'
 test_unit_test_SOURCES = \
 	test/test_aestag.c \
 	test/test_cipher.c \
--- a/test/test_ecc.c
+++ b/test/test_ecc.c
@@ -1411,7 +1411,7 @@ int test_ec_load_key(void* data)
     params[0] = OSSL_PARAM_construct_end();
 
     PRINT_MSG("Open ECC private key");
-    ctx = OSSL_STORE_open_ex(CERTS_DIR "ecc-key.pem", wpLibCtx, NULL, NULL, NULL,
+    ctx = OSSL_STORE_open_ex(CERTS_DIR "/ecc-key.pem", wpLibCtx, NULL, NULL, NULL,
         params, NULL, NULL);
     err = ctx == NULL;
     if (err == 0) {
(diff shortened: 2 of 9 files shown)


bandi13:
retest this please

test/unit.h
```diff
@@ -62,6 +62,10 @@ void print_buffer(const char *desc, const unsigned char *buffer, size_t len);
 #define TEST_DECL(func, data)        { #func, func, data, 0, 0, 0 }
 #endif
 
+#ifndef CERTS_DIR
+#define CERTS_DIR "./certs/"
```
douzzer, line 66: directory path fragments in vars/macros shouldn't include the trailing slash -- the slashes should be added in the course of assembling the full path.  that way, bad values for the var/macro lead to immediate "not a directory" errors, rather than mysterious later failures.  it's also more readable code.

btw, on Linux and BSD at least, multiple slashes work like one:
```
$ ls -l /bin/////ls
-r-xr-xr-x  1 root wheel 34408 Nov 10  2023 /bin/////ls
```

test/test_ecc.c
```diff
@@ -1411,7 +1411,7 @@ int test_ec_load_key(void* data)
     params[0] = OSSL_PARAM_construct_end();
 
     PRINT_MSG("Open ECC private key");
-    ctx = OSSL_STORE_open_ex("./certs/ecc-key.pem", wpLibCtx, NULL, NULL, NULL,
+    ctx = OSSL_STORE_open_ex(CERTS_DIR"ecc-key.pem", wpLibCtx, NULL, NULL, NULL,
```
douzzer, line 1414: see below note.  should be `OSSL_STORE_open_ex(CERTS_DIR "/ecc-key.pem", ...)`.  note also it's fine to put whitespace between `CERTS_DIR` and `"/ecc-key.pem"` -- makes for clearer code, and is ignored by the compiler.

test/test_ecc.c
```diff
@@ -1446,7 +1446,7 @@ int test_ec_load_cert(void* data)
     params[0] = OSSL_PARAM_construct_end();
 
     PRINT_MSG("Open certificate with ECC public key");
-    ctx = OSSL_STORE_open_ex("./certs/server-ecc.pem", wpLibCtx, NULL, NULL,
+    ctx = OSSL_STORE_open_ex(CERTS_DIR"server-ecc.pem", wpLibCtx, NULL, NULL,
```
douzzer, line 1449: tweak as above

test/test_rsa.c
```diff
@@ -749,7 +749,7 @@ int test_rsa_load_key(void* data)
     params[0] = OSSL_PARAM_construct_end();
 
     PRINT_MSG("Open RSA private key");
-    ctx = OSSL_STORE_open_ex("./certs/server-key.pem", wpLibCtx, NULL, NULL,
+    ctx = OSSL_STORE_open_ex(CERTS_DIR"server-key.pem", wpLibCtx, NULL, NULL,
```
douzzer, line 752: tweak as above

test/test_rsa.c
```diff
@@ -784,7 +784,7 @@ int test_rsa_load_cert(void* data)
     params[0] = OSSL_PARAM_construct_end();
 
     PRINT_MSG("Open certificate with RSA public key");
-    ctx = OSSL_STORE_open_ex("./certs/server-cert.pem", wpLibCtx, NULL, NULL,
+    ctx = OSSL_STORE_open_ex(CERTS_DIR"server-cert.pem", wpLibCtx, NULL, NULL,
```
douzzer, line 787: tweak as above

test/include.am
```diff
@@ -9,6 +9,8 @@ check_PROGRAMS += test/unit.test
 noinst_PROGRAMS += test/unit.test
 DISTCLEANFILES += test/.libs/unit.test
 
+# Override the 'certs' directory with the toplevel source directory for 'make distcheck' to work
+test_unit_test_CPPFLAGS  = -DCERTS_DIR='"$(abs_top_srcdir)/certs/"'
```
douzzer, line 13: omit trailing slash (see below)

douzzer (changes requested):

test/unit.h
bandi13: The trailing slash is on there, because if "CERTS_DIR" is not defined, I wanted the path to be in the current working directory. Otherwise the path would be at the root directory which would definitely cause issues.

bandi13: Additionally, I would argue `CERTS_DIR "/filename.ext"` is more readable than `CERTS_DIR "filename.ext"`. The missing slash would be an eye sore for me.

douzzer (approved):

PR #42, opened 2024-07-30
Update scripts

bandi13:
Adding in a utils-general.sh that will define required functions that can be overwritten by calling scripts. This way, there is a default definition without the user needing to define one.

Also have each module append paths to LD_LIBRARY_PATH as needed.


SparkiDev:
Put in some kind of description please.

SparkiDev (approved):

dgarske (approved):

PR #46, opened 2024-08-21
Fix settable func

bandi13:
Add in fixes for FIPS mode execution and testing scripts. Explicitly declared the parameters supported by the functions for HMAC, HKDF, and TLS1_PRF.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,8 +53,11 @@
 *.gcda
 
 # Installed dependencies
-/openssl*
-/wolfssl*
+/openssl-source/
+/openssl-install/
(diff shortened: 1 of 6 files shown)


xnox (approved):
This looks appropriate.

Previous code seemed to rely on default provider being loaded and looking over it's shoulder to see which settings the other provider offers.

ColtonWilley:
While this is a small duplication of settable params, current code is wrong so I say lets merge this as is and I will modify it later.

ColtonWilley (approved):

PR #54, opened 2024-10-09
Fix builds

bandi13:
New 'wc_HashAlg' structure now wraps the type into the structure as well as the union of the hash data structure. This caused some problems with the way wolfProvider was using it.


bandi13:
retest this please

douzzer (approved):

PR #56, opened 2024-11-06
Easier to use system's openssl library to clone

bandi13:


bandi13:
retest this please

douzzer:
couple failures -- checked one, it was `unit.test`, but no hints in the log as to what failed exactly.

bandi13:
> couple failures -- checked one, it was `unit.test`, but no hints in the log as to what failed exactly.

Yea, this is what I talked to you about before. Should have assigned it back to me. Still debugging.

douzzer (approved):

PR #63, opened 2025-01-24
Initial implementation of RSA X931 for wolfProvider

ColtonWilley:


src/wp_rsa_sig.c
```diff
@@ -929,6 +1112,163 @@ static int wp_rsa_verify_no_pad(wp_RsaSigCtx* ctx, const unsigned char* sig,
...
+        if ((decryptedSig[sigLen-1] & 0x0F) != 12) {
+            rc = mp_init_multi(&toMp, &nMinusTo, NULL, NULL, NULL, NULL);
+            if (rc != MP_OKAY) {
+                ok = 0;
+            }
+            if (ok) {
+                rc = mp_read_unsigned_bin(&toMp, decryptedSig, (int)sigLen);
+                if (rc != MP_OKAY) {
+                    ok = 0;
+                } else {
```
SparkiDev, line 1224: Formatting:
```
}
else {
```

SparkiDev (changes requested):
Please add tests as well.

src/wp_rsa_sig.c
ColtonWilley: Done

SparkiDev (approved):

src/wp_rsa_sig.c
```diff
@@ -651,6 +653,56 @@ static int wp_rsa_sign_pss(wp_RsaSigCtx* ctx, unsigned char* sig,
...
+ * Add X9.31 padding to the input buffer, placing the result in the output
+ * buffer.
+ *
+ * @param  to       [out]  Buffer to store padded result.
+ * @param  toLen    [in]   Length of "to" buffer.
+ * @param  from     [in]   Input buffer.
+ * @param  fromLen  [in]   Length of input buffer.
+ * @returns  1 on success and 0 on failure.
+ */
+static int wp_add_x931_padding(unsigned char* to, size_t toLen,
```
dgarske, line 666: @ColtonWilley should this padding scheme be added to wolfCrypt also (or even instead of being here)?

dgarske (changes requested):

src/wp_rsa_sig.c
ColtonWilley: I talked this over with Sean and he was in favor of having it at the wolfProvider layer, analogous to how wolfEngine has it.

SparkiDev: If we add it to wolfSSL, older FIPS libraries won't support it.
So we need this code for older FIPS libraries.

dgarske (approved):

PR #77, opened 2025-03-04
Fix RSA import use case for python

ColtonWilley:


ColtonWilley:
Retest this please.

src/wp_rsa_kmgmt.c
```diff
@@ -961,6 +961,15 @@ static int wp_rsa_validate(const wp_Rsa* rsa, int selection, int checkType)
     return ok;
 }
 
+static int numParams(const OSSL_PARAM *p) {
```
SparkiDev, line 964: Format and function rename:
```
static int wp_params_count(const OSSL_PARAM *p)
{
    int cnt = 0;
    while ((p != NULL) && (p->key != NULL)) {
         cnt++;
         p++;
     }
     return cnt;
}
```

SparkiDev (changes requested):

SparkiDev (approved):

PR #80, opened 2025-03-06
Fix param handling for ec encoding for openssl genpkey flow

ColtonWilley:
Add aliases to the curve table for "P-*" representations of the curve. Allow explicit setting of encoding for named_curve on keygen.

### DIFF
--- a/src/wp_ecc_kmgmt.c
+++ b/src/wp_ecc_kmgmt.c
@@ -116,10 +116,15 @@ static int wp_ecc_gen_set_params(wp_EccGenCtx* ctx, const OSSL_PARAM params[]);
 /** Mapping of OpenSSL curve name to wolfSSL elliptic curve information. */
 static const wp_EccGroupMap wp_ecc_group_map[] = {
     { SN_X9_62_prime192v1, ECC_SECP192R1, 192 },
+    { "P-192"            , ECC_SECP192R1, 192 },
     { SN_secp224r1       , ECC_SECP224R1, 224 },
+    { "P-224"            , ECC_SECP224R1, 224 },
     { SN_X9_62_prime256v1, ECC_SECP256R1, 256 },
+    { "P-256"            , ECC_SECP256R1, 256 },
     { SN_secp384r1       , ECC_SECP384R1, 384 },
+    { "P-384"            , ECC_SECP384R1, 384 },
     { SN_secp521r1       , ECC_SECP521R1, 521 },
+    { "P-521"            , ECC_SECP521R1, 521 },
 };
 
 /** Number of entries in elliptic curve mapping. */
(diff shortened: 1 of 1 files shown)


src/wp_ecc_kmgmt.c
```diff
@@ -1590,6 +1598,19 @@ static int wp_ecc_gen_set_params(wp_EccGenCtx* ctx, const OSSL_PARAM params[])
         ctx->curveName, sizeof(ctx->curveName)))) {
         ok = 0;
     }
+    if (ok) {
+        p = OSSL_PARAM_locate_const(params,
+                OSSL_PKEY_PARAM_EC_ENCODING);
+        if ((p != NULL) && ((p->data_type != OSSL_PARAM_UTF8_STRING) ||
+                            (p->data_size != \
```
SparkiDev, line 1605: remove line continuation character.

src/wp_ecc_kmgmt.c
```diff
@@ -1590,6 +1598,19 @@ static int wp_ecc_gen_set_params(wp_EccGenCtx* ctx, const OSSL_PARAM params[])
         ctx->curveName, sizeof(ctx->curveName)))) {
         ok = 0;
     }
+    if (ok) {
+        p = OSSL_PARAM_locate_const(params,
+                OSSL_PKEY_PARAM_EC_ENCODING);
+        if ((p != NULL) && ((p->data_type != OSSL_PARAM_UTF8_STRING) ||
```
SparkiDev, line 1604: Messy.
Check for not NULL and then if else if else if on the || checks.

src/wp_ecc_kmgmt.c
```diff
@@ -1590,6 +1598,19 @@ static int wp_ecc_gen_set_params(wp_EccGenCtx* ctx, const OSSL_PARAM params[])
         ctx->curveName, sizeof(ctx->curveName)))) {
         ok = 0;
     }
+    if (ok) {
+        p = OSSL_PARAM_locate_const(params,
+                OSSL_PKEY_PARAM_EC_ENCODING);
```
SparkiDev, line 1603: Encodings are uncompressed, compressed and hybrid.

SparkiDev (changes requested):
Add test

src/wp_ecc_kmgmt.c
ColtonWilley: I believe you have this confused with param OSSL_PKEY_PARAM_EC_POINT_CONVERSION_FORMAT, which has values uncompressed, compressed and hybrid. This param is OSSL_PKEY_PARAM_EC_ENCODING, which has the values named_curve and explicit. See the tables at the top of openssl/crypto/ec/ec_backend.c for details.

src/wp_ecc_kmgmt.c
ColtonWilley: Reworked it as suggested.

SparkiDev:
Closed in favour of #90.
Pushing to this PR wasn't working.

PR #83, opened 2025-03-11
Fix HKDF handling to allow setting NULL/0 salt

ColtonWilley:


src/wp_hkdf.c
```diff
@@ -304,12 +305,14 @@ static int wp_hkdf_base_set_ctx_params(wp_HkdfCtx* ctx,
         if (ok && (!wp_hkdf_base_get_mode(params, &ctx->mode))) {
             ok = 0;
         }
-        if (ok && (!wp_params_get_octet_string(params, OSSL_KDF_PARAM_KEY,
-                &ctx->key, &ctx->keySz, 1))) {
+        p = OSSL_PARAM_locate((OSSL_PARAM *)params, OSSL_KDF_PARAM_KEY);
```
SparkiDev, line 308: This lookup is being done now even when !ok.
And the getting of the octet string is doing the lookup again.

SparkiDev (changes requested):

src/wp_hkdf.c
ColtonWilley: I new it was a double lookup but thought that was cleaner than duplicating the code from inside wp_params_get_octet_string. I have changed it to remove double lookup.

PR #109, opened 2025-04-15
Fix ECC issue where y component was incorrectly overwritten

ColtonWilley:
Fix issue found when encoding where wp_ecc_set_params_x_y() incorrectly overwrites a previously decoded value for the public y component. 

Test case involves pkcs11-tool interop testing with openssl, will be brought in with pending OpenSC CI for wolfProvider.


ColtonWilley:
This fix was reworked and brought in as part of https://github.com/wolfSSL/wolfProvider/pull/115.

PR #120, opened 2025-04-25
Temporary test fix due to changing behavior in openssl master

ColtonWilley:
This PR for openssl https://github.com/openssl/openssl/pull/27305 went into master, changing expected return codes. Only run tests for older versions until this can be worked out with the openssl maintainers.


padelsbach:
LGTM. Any more details on what the specific incompatibility is?

ColtonWilley:
> LGTM. Any more details on what the specific incompatibility is?

Yes, previous behavior that allowed setting a salt of NULL/0 has been changed to now error on that case.

dgarske (approved):

PR #123, opened 2025-04-30
Github workflow testing support with `WOLFPROV_FORCE_FAIL=1`

aidangarske:
# Description

- Adds support for WPFF matrix refs for `sssd.yml`, `nginx.yml`, `openvpn.yml`, `curl.yml`, `simple.yml`, and `stunnel.yml`
- Updated `check-workflow-result.sh` to pass in sssd, stunnel, nginx, curl, and openvpn as test_suite
- Injects the WPFF right before we do make check and preserves original workflow functionality. 
- Expanded `sssd.yml` test so we can inject the WPFF ref after we make and before make check
- use `ossl-depends-${{ matrix.openssl_ref }}` so we can supply different openssl refs without breaking all the other builds when caching and retrieving. 
- Only test master with default build if WPFF we just neg test stable version

### DIFF
--- a/.github/workflows/curl.yml
+++ b/.github/workflows/curl.yml
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         wolfssl_ref: [ 'master', 'v5.7.4-stable' ]
+        openssl_ref: [ 'openssl-3.2.0' ]
     steps:
       - name: Checkout wolfProvider
         uses: actions/checkout@v4
@@ -49,7 +50,7 @@ jobs:
             openssl-source
             openssl-install
 
-          key: ossl-depends
+          key: ossl-depends-${{ matrix.openssl_ref }}
 
       # If not yet built this version, build it now
       - name: Build wolfProvider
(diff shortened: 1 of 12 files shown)


.github/workflows/curl.yml
```diff
@@ -73,7 +74,11 @@ jobs:
       matrix:
         curl_ref: [ 'master', 'curl-8_4_0' ]
         wolfssl_ref: [ 'master', 'v5.7.4-stable' ]
+        openssl_ref: [ 'openssl-3.2.0' ]
```
padelsbach, line 77: Any reason to not test newer versions of openssl? Seems that prior to this change, we tested openssl master right? Any reason to not test that?

.github/workflows/curl.yml
```diff
@@ -73,7 +74,11 @@ jobs:
...
         force_fail: ['WOLFPROV_FORCE_FAIL=1', '']
+        exclude:
```
padelsbach, line 79: Just curious, does this mean we don't test WPFF with curl master branch specifically? Why is that needed?

ColtonWilley: My thought is that the force fail testing is very sensitive to changes, a single change to their test framework and now our entire CI is held up on it. I am thinking only doing force fail testing against latest release is a better way to handle it.

.github/workflows/curl.yml
aidangarske: Prior to these changes we tested `openssl-3.2.0` just like we are now. This is because `./scripts/build-wolfprovider.sh` builds 3.2.0 by default. My thinking with adding the `openssl_ref` was that we can now test with different versions of openssl if we wanted to like you suggest. I don't see why we couldn't test newer versions. Though I have not manually verified that any of the osp repos work with newer openssl versions yet...

padelsbach: Makes sense, thanks. I suppose we can test against newer versions of openssl in a future PR

padelsbach (approved):

PR #124, opened 2025-05-01
Add test cases for EVP_PKEY_fromdata, fix incompatibility in wp_rsa_mport_key_data().

padelsbach:


src/wp_rsa_kmgmt.c
```diff
@@ -1087,28 +1080,43 @@ static int wp_rsa_import_key_data(wp_Rsa* rsa, const OSSL_PARAM params[],
...
-            if (!wp_mp_read_unsigned_bin_le(mp, p->data, p->data_size)) {
+    if (ok) {
+        int cnt = wp_params_count(params);
+        rsa->key.type = priv ? RSA_PRIVATE : RSA_PUBLIC;
+
+        for (int i = 0; i < cnt; i++) {
+            /* Use the table to look up the offset in the rsa struct */
+            const OSSL_PARAM* p = &params[i];
+            int index = -1;
+            for (unsigned j = 0; j < ARRAY_SIZE(wp_rsa_param_key); j++) {
```
ColtonWilley, line 1100: All variable declarations at start of function.

src/wp_rsa_kmgmt.c
```diff
@@ -1087,28 +1080,43 @@ static int wp_rsa_import_key_data(wp_Rsa* rsa, const OSSL_PARAM params[],
...
+                if (XSTRNCMP(p->key, wp_rsa_param_key[j], p->data_size) == 0) {
+                    index = j; 
+                    break;
+                }
+            }
+            if (index < 0) {
```
ColtonWilley, line 1106: Please investigate if openssl fails with erroneous/extra params in the list. If extra params are treated as a noop for them we should do the same.

ColtonWilley (changes requested):

src/wp_rsa_kmgmt.c
padelsbach: The Wolf coding standard doc seems to indicate this is ok. But sure, easy enough to change.

src/wp_rsa_kmgmt.c
padelsbach: Good catch. OSSL does not fail with extra params. I will add a test case and fix.

ColtonWilley (approved):

PR #125, opened 2025-05-01
Only restore ed key clamping if key was imported

ColtonWilley:
Libssh2 generates and then exports a curve25519 private key. If we generated the key, we are incorrectly clamping the private key with zero values on export. Fix to only clamp if key was imported.

### DIFF
--- a/src/wp_ecx_kmgmt.c
+++ b/src/wp_ecx_kmgmt.c
@@ -168,6 +168,8 @@ struct wp_Ecx {
     unsigned int hasPub:1;
     /** Private key available. */
     unsigned int hasPriv:1;
+    /* Private key was imported and has been clamped */
+    unsigned int clamped:1;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


padelsbach:
`ecx->hasPriv` is checked for private keys around like 793. Should `ecx->clamped` be checked there as well?

ColtonWilley:
> `ecx->hasPriv` is checked for private keys around like 793. Should `ecx->clamped` be checked there as well?

I dont think so. Whether it has an underlying private key is independent of clamping.

padelsbach:
> `ecx->hasPriv` is checked for private keys around like 793. Should `ecx->clamped` be checked there as well?
> 
> Edit: I believe not since this key may have been generated, and thus not clamped.

In any case, it might be good to add a comment that the `clamped` flag is specifically for import and not generated.

SparkiDev (approved):

PR #142, opened 2025-05-19
Add Cache Key for Openssl Version Cache / Restore

aidangarske:
# Description

- Fixes version inconsistency error with openssl 
- Uses a cache key like wolfssl so we get a clean build every time we invoke a test workkflow


padelsbach (approved):
Thanks for the quick fix

douzzer (approved):

PR #150, opened 2025-05-28
Rewrite AES-GCM stream handling for FIPS

ColtonWilley:
The current AES-GCM implementation when FIPS is enabled does not have underlying stream support. This PR modifies wolfProvider to properly meet streaming requirements when the underlying AES-GCM streaming APIs are not available.

### DIFF
--- a/src/wp_aes_aead.c
+++ b/src/wp_aes_aead.c
@@ -1337,8 +1337,6 @@ static int wp_aesgcm_stream_final(wp_AeadCtx *ctx, unsigned char *out,
  * @param [in, out] ctx      AEAD context object.
  * @param [out]     out      Buffer to hold decrypted/encrypted data.
  * @param [out]     outLen   Length of data in output buffer.
- * @param [in]      in       Data to be encrypted/decrypted.
- * @param [in]      inLen    Length of data to be encrypted/decrypted.
  * @param [in]      offset   Offset into the spooled data buffer.
  * @param [in]      done     This is a final operation.
  *
--- a/src/wp_aes_aead.c
+++ b/src/wp_aes_aead.c
@@ -32,6 +32,16 @@
 
 #if defined(WP_HAVE_AESGCM) || defined(WP_HAVE_AESCCM)
 
+/* For non-streaming AES-GCM, we have to spool all previous updates.
+ * This is the number of extra bytes we add when allocating the internal
+ * buffer used to spool the stream input. This way we if there is space
+ * we can use the existing buffer, reducing the number of full realloc + copy
+ * operations we need to do. Increase this number for better performance and
+ * more memory usage, decrease for worse performance but less overhead */
+#ifndef WP_AES_GCM_EXTRA_BUF_LEN
+#define WP_AES_GCM_EXTRA_BUF_LEN 128
+#endif
+
 /**
  * Authenticated Encryption with Associated Data structure.
  */
@@ -94,10 +104,14 @@ typedef struct wp_AeadCtx {
     /** CCM is not streaming and needs to cache AAD data. */
     unsigned char* aad;
 #if defined(WP_HAVE_AESGCM) && !defined(WOLFSSL_AESGCM_STREAM)
-    /** Length of AAD data cached.  */
+    /** Length of data cached.  */
     size_t inLen;
     /** CCM is not streaming and needs to cache AAD data. */
     unsigned char* in;
+    /* Total buffer size */
+    size_t bufSize;
+    /* Original IV */
+    unsigned char origIv[AES_BLOCK_SIZE];
 #endif
 } wp_AeadCtx;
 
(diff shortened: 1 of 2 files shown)


src/wp_aes_aead.c
```diff
@@ -1414,12 +1474,22 @@ static int wp_aesgcm_stream_update(wp_AeadCtx *ctx, unsigned char *out,
             ok = 0;
         }
         else if (inLen > 0) {
+            curLen = ctx->inLen;
```
padelsbach, line 1477: Seems like we should cache the data before setting `curLen` to ensure that all cached data is processed in `wp_aesgcm_encdec` below, right?

Or is `inLen` just the current chunk?

src/wp_aes_aead.c
```diff
@@ -1312,65 +1339,96 @@ static int wp_aesgcm_stream_final(wp_AeadCtx *ctx, unsigned char *out,
...
 
     if (ctx->tagLen == UNINITIALISED_SIZET) {
         ctx->tagLen = EVP_GCM_TLS_TAG_LEN;
     }
 
-    if (ctx->enc) {
-        if (!ctx->ivSet) {
-            rc = wc_AesGcmSetExtIV(&ctx->aes, ctx->iv, (word32)ctx->ivLen);
-            if (rc != 0) {
+    if (ctx->inLen > offset || (ctx->tagAvail == 0 && done)) {
```
padelsbach, line 1360: Should this be `>=` to offset? Does it work when `ctx->inLen` is 0 (no data cached) and offset is some number of bytes? (Maybe that scenario does not take this path?)

src/wp_aes_aead.c
```diff
@@ -1414,12 +1474,22 @@ static int wp_aesgcm_stream_update(wp_AeadCtx *ctx, unsigned char *out,
...
             }
+            else {
+                process = 1;
+            }
         }
 
-        *outLen = oLen;
+        /* If there is data to process, do it now */
+        if (process) {
+            ok = wp_aesgcm_encdec(ctx, out, outLen, curLen, 0);
```
padelsbach, line 1488: Should we flush the cache after this since the caller has retrieved the data?

test/test_aestag.c
```diff
@@ -363,6 +414,57 @@ static int test_aes_tag_fixed_enc(const EVP_CIPHER *cipher,
     return err;
 }
 
+static int test_aes_tag_fixed_enc_multi(const EVP_CIPHER *cipher,
```
padelsbach, line 417: Nice add of the `*_multi` test cases!

src/wp_aes_aead.c
```diff
@@ -1312,65 +1339,96 @@ static int wp_aesgcm_stream_final(wp_AeadCtx *ctx, unsigned char *out,
...
+        /* Prepare a temp buffer to store all the output */
+        if (ctx->inLen > 0) {
+            tmp = OPENSSL_zalloc(ctx->inLen);
+            if (tmp == NULL) {
                 ok = 0;
             }
         }
+        /* Once loaded, always use original IV */
+        iv = ctx->iv;
+        if (ctx->ivState == IV_STATE_COPIED) {
```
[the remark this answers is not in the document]
ColtonWilley: Honestly its a bit complicated, the semantics are replicated from the openssl default provider internals. Buffered vs copied has to do with fixed vs random portion of IVs used in TLS, sometimes we just take what we get on init, sometimes we actually generate a portion, or set a fixed piece in addition to what we got on init. Either way all paths flow to "copied state means we are done", which guarantees we now properly reuse the original IV.

src/wp_aes_aead.c
ColtonWilley: No, we need to spool the entire message as we go otherwise we cant compute the correct tag value on an additional update.

src/wp_aes_aead.c
```diff
@@ -1312,65 +1339,96 @@ static int wp_aesgcm_stream_final(wp_AeadCtx *ctx, unsigned char *out,
...
  * @param [in]      in       Data to be encrypted/decrypted.
  * @param [in]      inLen    Length of data to be encrypted/decrypted.
+ * @param [in]      offset   Offset into the spooled data buffer.
+ * @param [in]      done     This is a final operation.
+ *
  * @return  1 on success.
  * @return  0 on failure.
  */
-static int wp_aesgcm_encdec(wp_AeadCtx *ctx, unsigned char *out, size_t* outLen)
+static int wp_aesgcm_encdec(wp_AeadCtx *ctx, unsigned char *out, size_t* outLen,
```
SparkiDev, line 1348: Function comments doesn't match function declaration.

SparkiDev (changes requested):
Otherwise looks good.

SparkiDev (approved):

PR #152, opened 2025-06-01
Add tests for DH get params and match OSSL behavior

padelsbach:
Added extensive test case as well.

### DIFF
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -685,21 +685,26 @@ static int wp_dh_get_params_encoded_public_key(wp_Dh* dh, OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY);
     if (p != NULL) {
-        size_t outLen = mp_unsigned_bin_size(&dh->key.p);
-
-        if (p->data != NULL) {
-            if (p->data_size < outLen) {
-                ok = 0;
-            }
-            if (ok) {
-                unsigned char* data = p->data;
-                size_t padSz = outLen - dh->pubSz;
-                /* Front pad with zeros. */
-                XMEMSET(data, 0, padSz);
-                XMEMCPY(data + padSz, dh->pub, dh->pubSz);
-            }
+        if (p->data_type != OSSL_PARAM_OCTET_STRING) {
+            ok = 0;
         }
-        p->return_size = outLen;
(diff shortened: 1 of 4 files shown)


ColtonWilley (approved):

src/wp_dh_kmgmt.c
```diff
@@ -719,39 +724,145 @@ static int wp_dh_get_params(wp_Dh* dh, OSSL_PARAM params[])
...
+        }
     }
-    if (ok && (!wp_params_set_mp(params, OSSL_PKEY_PARAM_FFC_G,
-                                 &dh->key.g, 1))) {
-        ok = 0;
+    if (ok) {
+        p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_FFC_P);
+        if (p != NULL) {
+            if (p->data == NULL) {
+                p->return_size = 256;
```
SparkiDev, line 782: Why 256? Do you mean to get the size of the mp_int?

SparkiDev (changes requested):

src/wp_dh_kmgmt.c
padelsbach: This follows the ossl behavior for `return_size`. Do we want to mimic it to this level of detail?

SparkiDev: Looking at OpenSSL. I cannot see 256 anywhere in dh_kmgmt.c.
I can see 2048 used for p bits but that is for generation.

SparkiDev (changes requested):

src/wp_dh_kmgmt.c
padelsbach: Ok done. I removed some of the strict checks on the sizes in the test code.

src/wp_dh_kmgmt.c
```diff
@@ -719,39 +724,125 @@ static int wp_dh_get_params(wp_Dh* dh, OSSL_PARAM params[])
...
+            else {
+                ok = 0;
+            }
+        }
+    }
+    if (ok) {
+        p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_PRIV_KEY);
+        if (p != NULL) {
+            if (p->data == NULL) {
+                p->return_size = 256;
```
SparkiDev, line 817: Fixed value here too.

SparkiDev (changes requested):

SparkiDev (approved):

PR #160, opened 2025-06-07
Draft: Add iperf workflow

padelsbach:


.github/workflows/iperf-files/rsa_private.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
padelsbach, line 1: I did receive a GitGuardian message about this 'secret' being added to the repo.

Ideally, we would generate this keypair as part of the test. However, it requires user input to set the passcode for the keypair. Please let me know if you see alternatives

padelsbach: Fixed this with the `-passin` and `-passout` options

PR #161, opened 2025-06-09
libwebsockets.yml Github workflow

BridgerVoss:
Added Github workflow for libwebsockets OSP
 - tests pass as expected

### DIFF
--- a/.github/scripts/check-workflow-result.sh
+++ b/.github/scripts/check-workflow-result.sh
@@ -255,6 +255,7 @@ if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
             fi
         else
             echo "Error: liboauth2-test.log not found"
+            exit 1
         fi
     # ----- TCPDUMP -----
     elif [ "$TEST_SUITE" = "tcpdump" ]; then
@@ -285,7 +286,7 @@ if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
                 echo "  Sent: $sender_gb GB, Received: $receiver_gb GB"
                 exit 1
             fi
-        else 
+        else
             echo "Error: $IPERF_TEST_LOG not found"
             exit 1
         fi
--- a/.github/workflows/libwebsockets.yml
+++ b/.github/workflows/libwebsockets.yml
@@ -1,0 +1,100 @@
+name: libwebsockets Tests
+
+# START OF COMMON SECTION
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+# END OF COMMON SECTION
+
+jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
+      wolfssl_ref: ${{ matrix.wolfssl_ref }}
+      openssl_ref: ${{ matrix.openssl_ref }}
+    strategy:
+      matrix:
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
+
+  test_libwebsockets:
+    runs-on: ubuntu-22.04
+    needs: build_wolfprovider
+    # This should be a safe limit for the tests to run.
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        libwebsockets_ref: [ 'main', 'v4.3.3' ]
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
+        force_fail: [ 'WOLFPROV_FORCE_FAIL=1', '' ]
+        exclude:
+          - libwebsockets_ref: 'main'
+            force_fail: 'WOLFPROV_FORCE_FAIL=1'
+    steps:
+      # Checkout the source so we can run the check-workflow-result script
+      - name: Checkout wolfProvider
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+              .github
+      - name: Retrieving wolfProvider from cache
+        uses: actions/cache/restore@v4
+        id: wolfprov-cache-restore
+        with:
+          path: |
+            scripts
+            wolfssl-source
+            wolfssl-install
+            wolfprov-install
+            openssl-install
+            provider.conf
+          key: wolfprov-${{ matrix.wolfssl_ref }}-${{ matrix.openssl_ref }}-${{ github.sha }}
+          fail-on-cache-miss: true
+
+      - name: Install libwebsockets dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc6 libcap2 zlib1g cmake build-essential
+      - name: Download libwebsockets
+        uses: actions/checkout@v4
+        with:
+          repository: warmcat/libwebsockets
+          ref: ${{ matrix.libwebsockets_ref }}
+          path: libwebsockets
+
+      - name: Build libwebsockets
+        working-directory: libwebsockets
+        run: |
+          mkdir build
+          cd build
+          cmake ..
+          make -j$(nproc)
+          sudo make install
+      - name: Run libwebsockets tests
+        working-directory: libwebsockets
+        run: |
+          echo "Setting environment variables..."
+          source $GITHUB_WORKSPACE/scripts/env-setup
+          export ${{ matrix.force_fail }}
+          ./build/bin/libwebsockets-test-server --port=11111 --ssl > server.log 2>&1 & SERVER_PID=$!
+          sleep 5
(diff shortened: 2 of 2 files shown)


.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,112 @@
...
+            wolfssl-install
+            wolfprov-install
+            provider.conf
+          key: wolfprov-${{ matrix.wolfssl_ref }}-${{ github.sha }}
+          fail-on-cache-miss: true
+
+      - name: Install libwebsockets dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc6 libcap2 libssl3 zlib1g cmake libuv1-dev build-essential
```
padelsbach, line 69: Is `libssl3` needed here? Seems like it might cause potential for using the wrong version. Maybe just `libssl-dev`?

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,112 @@
...
+        run: |
+          set -e
+
+          echo "Setting environment variables..."
+          export LD_LIBRARY_PATH=$GITHUB_WORKSPACE/wolfssl-install/lib:$GITHUB_WORKSPACE/openssl-install/lib64
+          export OPENSSL_CONF=$GITHUB_WORKSPACE/provider.conf
+          export OPENSSL_MODULES=$GITHUB_WORKSPACE/wolfprov-install/lib
+          export ${{ matrix.force_fail }}
+
+          ./build/bin/libwebsockets-test-server --port=11111 --ssl > server.log 2>&1 & SERVER_PID=$!
```
padelsbach, line 98: Can we verify that the 2 executables and openssl itself are using provider? Something like this: 

```
./build/bin/libwebsockets-test-server | grep wolfProvider
./build/bin/libwebsockets-test-client | grep wolfProvider
$GITHUB_WORKSPACE/openssl-install/bin/openssl list -providers | grep libwolfprov
```

.github/scripts/check-workflow-result.sh
```diff
@@ -270,8 +270,21 @@ if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
                 echo "  Sent: $sender_gb GB, Received: $receiver_gb GB"
                 exit 1
             fi
-        else 
+        else
             echo "Error: $IPERF_TEST_LOG not found"
+    # ----- LIBWEBSOCKETS -----
+    elif [ "$TEST_SUITE" = "libwebsockets" ]; then
+        if [ -f "libwebsockets-test.log" ]; then
```
padelsbach, line 277: Consider putting `libwebsockets-test.log` in a variable for DRYness

.github/scripts/check-workflow-result.sh
```diff
@@ -270,8 +270,21 @@ if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
                 echo "  Sent: $sender_gb GB, Received: $receiver_gb GB"
                 exit 1
             fi
-        else 
+        else
             echo "Error: $IPERF_TEST_LOG not found"
```
padelsbach, line 274: There seems to be a missing end to this `else` case. I think there should be the following: 
```
  exit 1
fi
```

padelsbach: Also, I know it's not part of your change, but can you please add `exit 1` around line 257?

aidangarske (changes requested):
Great job!

aidangarske (dismissed):

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,121 @@
...
+            wolfssl-install
+            wolfprov-install
+            provider.conf
+          key: wolfprov-${{ matrix.wolfssl_ref }}-${{ github.sha }}
+          fail-on-cache-miss: true
+
+      - name: Install libwebsockets dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc6 libcap2 libssl-dev zlib1g cmake build-essential
```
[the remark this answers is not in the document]
aidangarske: We should not need `libssl-dev` Paul is right. Please try without `libssl-dev`.

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,124 @@
...
+
+          echo "Setting environment variables..."
+          export LD_LIBRARY_PATH=$GITHUB_WORKSPACE/wolfssl-install/lib:$GITHUB_WORKSPACE/openssl-install/lib64
+          export OPENSSL_CONF=$GITHUB_WORKSPACE/provider.conf
+          export OPENSSL_MODULES=$GITHUB_WORKSPACE/wolfprov-install/lib
+          export ${{ matrix.force_fail }}
+
+          ./build/bin/libwebsockets-test-server --port=11111 --ssl > server.log 2>&1 & SERVER_PID=$!
+          sleep 5
+          set +e
```
aidangarske, line 103: Do we really need all these `set -e` and `set +e`? Are these commands not exiting when they fail?

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,121 @@
...
+            wolfssl-install
+            wolfprov-install
+            provider.conf
+          key: wolfprov-${{ matrix.wolfssl_ref }}-${{ github.sha }}
+          fail-on-cache-miss: true
+
+      - name: Install libwebsockets dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc6 libcap2 libssl3 zlib1g cmake build-essential
```
padelsbach, line 72: Is `libssl3` required here? I think it's best to avoid having another copy of ssl on the system.

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,121 @@
...
+
+          cat client.log || echo "Missing client.log"
+
+          cat server.log client.log > libwebsockets-test.log
+          set +e
+          grep -q "error:03080006" libwebsockets-test.log || grep -q "Failed to create default vhost" libwebsockets-test.log;
+          set -e
+          TEST_RESULT=$((1 - $?))
+
+          bash -x $GITHUB_WORKSPACE/.github/scripts/check-workflow-result.sh $TEST_RESULT ${{ matrix.force_fail }} libwebsockets
```
padelsbach, line 121: Why use `bash -x` here?

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,121 @@
...
+          grep libwolfprov provider-list.log || (echo "ERROR: libwolfprov not found in OpenSSL providers" && exit 1)
+
+          kill $SERVER_PID || echo "Server already exited"
+
+          cat server.log || echo "Missing server.log"
+
+          cat client.log || echo "Missing client.log"
+
+          cat server.log client.log > libwebsockets-test.log
+          set +e
```
padelsbach, line 116: Do we need the `set +e` and `set -e` here?

padelsbach (commented):
Looking good. A few more questions. And it looks like the test is failing, can you investigate?

.github/workflows/libwebsockets.yml
BridgerVoss: ill get this fixed I was debugging earlier and forgot to remove this.

.github/workflows/libwebsockets.yml
BridgerVoss: I updated it. Since Paul suggested I can check for failures in yml file and now need no special handling in check-workflow script if my grep fails then my whole test fails for some reason even without set -e beforehand. Now this is only spot that needs set +e and set -e is around the grep command.

.github/workflows/libwebsockets.yml
padelsbach: Ah I see. Another approach I've used is something like this. It allows `<command>` to fail, but also capture the test result without exiting the test:

```
          if <command>; then
            TEST_RESULT=1
          else
            TEST_RESULT=0
          fi
```

padelsbach (dismissed):
Nicely done!

aidangarske (dismissed):

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,123 @@
...
+          - libwebsockets_ref: 'main'
+            force_fail: 'WOLFPROV_FORCE_FAIL=1'
+    steps:
+      # Checkout the source so we can run the check-workflow-result script
+      - name: Checkout wolfProvider
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+              .github
+      - name: Retrieving OpenSSL from cache
```
padelsbach, line 47: There is a new caching mechanism on master, and openssl does not need to be separately retrieved. Can you please update based on some of the examples on master?

.github/workflows/libwebsockets.yml
```diff
@@ -0,0 +1,123 @@
...
+          mkdir build
+          cd build
+          cmake ..
+          make
+          sudo make install
+
+      - name: Run libwebsockets tests
+        working-directory: libwebsockets
+        run: |
+          echo "Setting environment variables..."
```
padelsbach, line 93: The env setup has also been updated/unified on master. Can you please update to follow that pattern?

padelsbach (approved):

PR #162, opened 2025-06-09
tcpdump wolfProvider test

JeremiahM37:
Added a yaml test file in GitHub workflows to test tcpdump compatibility with wolfProvider.
All tests pass and it seems to work with wolfProvider just fine.

### DIFF
--- a/.github/workflows/tcpdump.yml
+++ b/.github/workflows/tcpdump.yml
@@ -1,0 +1,126 @@
+name: tcpdump Tests
+
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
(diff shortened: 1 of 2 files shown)


.github/workflows/tcpdump.yml
```diff
@@ -0,0 +1,126 @@
...
+          export OPENSSL_MODULES="$GITHUB_WORKSPACE/wolfprov-install/lib"
+          export PKG_CONFIG_PATH="$GITHUB_WORKSPACE/openssl-install/lib64/pkgconfig:$GITHUB_WORKSPACE/libpcap-install/lib/pkgconfig:$PKG_CONFIG_PATH"
+          export PATH="$GITHUB_WORKSPACE/tcpdump-install/sbin:$GITHUB_WORKSPACE/tcpdump-install/bin:$PATH"
+          
+          # Apply the force_fail matrix variable
+          export ${{ matrix.force_fail }}
+
+          # Debug: Check if wolfProvider is loading correctly
+          echo "Testing wolfProvider availability:"
+          openssl list -providers || true
```
padelsbach, line 113: We need to check that our provider is listed in this command, and not mask any failure with `|| true`.

Something like this: 
```
$GITHUB_WORKSPACE/openssl-install/bin/openssl list -providers | grep libwolfprov
```

padelsbach (commented):
Looking good, just a couple minor questions

PR #166, opened 2025-06-11
Verify that wolfProvider loads for all workflows

aidangarske:
# Description 

- Verify that wolfProvider is loaded for each workflow 
- Update IPMITOOL to consistent version not commit tag

### DIFF
--- a/.github/workflows/cjose.yml
+++ b/.github/workflows/cjose.yml
@@ -2,7 +2,7 @@ name: cjose Tests
 
 # START OF COMMON SECTION
 on:
-  push: 
+  push:
     branches: [ 'master', 'main', 'release/**' ]
   pull_request:
     branches: [ '*' ]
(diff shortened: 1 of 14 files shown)


padelsbach (dismissed):
Looks good, nice improvement

.github/workflows/cjose.yml
```diff
@@ -98,8 +98,9 @@ jobs:
           export ${{ matrix.force_fail }}
           export SHELL=/bin/bash
 
-          # Verify OpenSSL loads wolfProvider
-          $GITHUB_WORKSPACE/openssl-install/bin/openssl list -providers | grep libwolfprov
+          echo "Checking OpenSSL providers:"
+          $GITHUB_WORKSPACE/openssl-install/bin/openssl list -providers | tee provider-list.log
+          grep libwolfprov provider-list.log || (echo "ERROR: libwolfprov not found in OpenSSL providers" && exit 1)
```
padelsbach, line 103: We should probably use `grep -q` to avoid extra output in the logs

aidangarske: Good point, let me add that.

PR #171, opened 2025-06-16
Fix test_ecx.c build if no ED448 in WolfSSL

adrianjarc:
If WolfSSL is built without `--enable-ed448` flag, the build of wolfProvider fails in `test/test_ecx.c` file because of undefined `ED448_KEY_SIZE` parameter.

This fix checks if ED25519 and ED448 are enabled and sets sizes for `readback_ossl` and `readback_wolf` accordingly.

### DIFF
--- a/src/wp_ecx_sig.c
+++ b/src/wp_ecx_sig.c
@@ -262,6 +262,32 @@ static int wp_ecx_digest_verify_init(wp_EcxSigCtx *ctx, const char *mdName,
     return ok;
 }
 
+/** Parameters that we support getting from the ECX signature context. */
+static const OSSL_PARAM wp_supported_gettable_ctx_params[] = {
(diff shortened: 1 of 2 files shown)


adrianjarc:
I am also now inspecting `src/wp_ecx_sig.c` file and see a few possible changes to take into account if ED448 or ED25519 are used. I will add them as a new commit to this PR.

adrianjarc:
@padelsbach I have added new changes that better handle definitions of ED25519 and ED448 specific functions in `src/wp_ecx_sig.c` file. Please review.

padelsbach:
I'm not able to reproduce the issue, perhaps mostly due to the fips build having several issues on aarch64.

But the change seems good to me. 

Do you have a contributor agreement in place with WolfSSL?

padelsbach (approved):

PR #177, opened 2025-06-17
tnftp.yml - Github CI Workflow

aidangarske:
# Description 

Adds support for github CI testing with tnftp version ref `tnftp-20210827` There is no testing suite for tnftp so I ran some simple operations that verify functionality and that WP is loaded and WPFF works as intended.

### DIFF
--- a/.github/scripts/check-workflow-result.sh
+++ b/.github/scripts/check-workflow-result.sh
@@ -41,11 +41,11 @@ fi
 if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
     # ----- CURL -----
     if [ "$TEST_SUITE" = "curl" ]; then
-        if [ -f "tests/curl-test.log" ]; then
+        if [ -f "curl-test.log" ]; then
             # Extract and clean the failed test list from the log
-            ACTUAL_FAILS=$(grep -a '^TESTFAIL: These test cases failed:' tests/curl-test.log | sed 's/.*failed: //')
+            ACTUAL_FAILS=$(grep -a '^TESTFAIL: These test cases failed:' curl-test.log | sed 's/.*failed: //')
         else
-            echo "Error: tests/curl-test.log not found"
+            echo "Error: curl-test.log not found"
             exit 1
         fi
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tnftp.yml
```diff
@@ -0,0 +1,107 @@
...
+              fi
+
+              # Test that tnftp can start (even if it fails to connect)
+              echo "Testing tnftp connection attempt..."
+              timeout 10 ./src/tnftp -n 192.0.2.1 2>&1 | head -10
+              echo "tnftp can attempt connections"
+
+              # Test SSL/TLS functionality
+              echo "Testing SSL/TLS connection..."
+              timeout 15 ./src/tnftp -n https://httpbin.org/get 2>&1
```
padelsbach, line 101: Looking better. With `WOLFPROV_FORCE_FAIL` does this timeout? If so, we might need to capture the failure more gracefully into `TEST_RESULT`

aidangarske: Looks like it doesn't time out and it retries a few times before failing. Looks like it fails pretty quick so we should be good.

padelsbach: Curious then what is the `timeout` for -- just to stop the test? Could the `-q <quittime>` option be used instead?

.github/workflows/tnftp.yml
```diff
@@ -0,0 +1,107 @@
...
+              if ./src/tnftp -? 2>&1 | grep -q "usage:"; then
+                  echo "tnftp help command works"
+              else
+                  echo "tnftp help command failed"
+                  exit 1
+              fi
+
+              # Test that tnftp can start (even if it fails to connect)
+              echo "Testing tnftp connection attempt..."
+              timeout 10 ./src/tnftp -n 192.0.2.1 2>&1 | head -10
```
padelsbach, line 96: I think we should test tnftp as a server as well. Recommend starting a server in daemon mode with the `-D` option, and then connect to it with the client, rather than connecting to httpbin. 

What do you think?

.github/workflows/tnftp.yml
aidangarske: Yes the timeout function just makes the test not run indefinitely if it doesn't work. I tried -q and it wasn't working like I wanted it to so I used timeout.

ColtonWilley (approved):

PR #180, opened 2025-06-17
Draft: Revert common caching strategy which was not working at scale

padelsbach:
There are a couple limitations with github actions....

First is that the caches aren't visible across branches -- with some exceptions if the cache was created on the main branch.

Second is that they don't guard against race conditions where multiple jobs access the cache. And the result is a corrupted cache entry or unexpected cache hit/miss

### DIFF
--- a/.github/workflows/build-wolfprovider.yml
+++ b/.github/workflows/build-wolfprovider.yml
@@ -12,140 +12,52 @@ on:
     outputs:
       cache_key:
         description: "Cache key for the build artifacts"
-        value: ${{ jobs.build_wolfprovider_common.outputs.cache_key }}
+        value: ${{ jobs.build_wolfprovider.outputs.cache_key }}
 
 jobs:
-  build_wolfprovider_common:
+  build_wolfprovider:
     name: Build wolfProvider
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     outputs:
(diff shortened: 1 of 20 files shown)


.github/workflows/build-wolfprovider.yml
```diff
@@ -12,141 +12,53 @@ on:
...
 
 jobs:
-  build_wolfprovider_common:
+  build_wolfprovider:
     name: Build wolfProvider
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     outputs:
-      cache_key: wolfprov-${{ inputs.wolfssl_ref }}-${{ inputs.openssl_ref }}-${{ github.sha }}
+      cache_key: wolfprov-${{ inputs.wolfssl_ref }}-${{ github.sha }}-${{ github.run_number }}
```
aidangarske, line 23: What is this `run_number` do we really need it? This is the only difference I see that might make a difference with the cache restore method

padelsbach: Github docs say it is:

> A unique number for each run of a particular workflow in a repository.

I had added this for testing to be sure that there were no collisions since the original way didn't work either.

aidangarske: Okay nice just making sure

PR #181, opened 2025-06-17
Modify FIPS build and fix AES-GCM stream handling for FIPS

ColtonWilley:
Until now wolfProvider has been building underlying wolfSSL FIPS using the same flags as default, including --enable-all-crypto. This is not a valid option for wolfSSL FIPS and it enables many algo flags in options.h that dont or should not have underlying crypto support. Fix build to use only needed flags.

Depending on build flags, wc_AesGcmDecrypt() implementation may or may not yield plaintext on a auth tag error. Modify stream handling to use wc_AesGcmEncrypt() instead to yield plaintext on all calls except the last.


SparkiDev (approved):
Looks correct to me.

SparkiDev (approved):

aidangarske (approved):

PR #185, opened 2025-06-20
Update Test Suite

ejohnstown:
When the check-workflow-result script runs, it grabs the commandline parameters into variables. If the force fail flag is left out, we are checking the force fail variable if it is blank, and then fixing it. We should check the test suite variable if blank and clear the force fail and update the test suite.


aidangarske (approved):
Good catch, thank you.

PR #190, opened 2025-06-23
Add `HAVE_GETPID` to options.h to fix `dh->ctx` override by rng

aidangarske:
# Description

- https://github.com/wolfSSL/wolfssl/pull/8867 introduced a problem with `HAVE_GETPID` that overrides the `dh->ctx` with the rng's value. Adding the `HAVE_GETPID` to options.h resolves these `WC_RNG rng` override issues. Specifically fixes the issues seen here on mac https://cloud.wolfssl-test.com/jenkins/job/wolfProvider/job/nightly-scripts-test/379/
- Also fixes oversight with the way we checkout our workflows now

### DIFF
--- a/.github/workflows/tnftp.yml
+++ b/.github/workflows/tnftp.yml
@@ -40,18 +40,16 @@ jobs:
         with:
           fetch-depth: 1
 
-      - name: Retrieving wolfSSL/wolfProvider from cache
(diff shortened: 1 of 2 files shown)


padelsbach (approved):
Good catch

PR #191, opened 2025-06-25
Implement FIPS support for testing scripts and fix FIPS AES cmd test

aidangarske:
# Description 

- Fixes FIPS AES cmd test failures with `WOLFPROV_DEBUG`
- Adds a `WOLFSSL_ISFIPS` env variable that gets passed in to testing scripts to enable FIPS mode testing. 
- Fixes bug in `./scripts/do-cmd-test.sh` with running AES-CFB with FIPS mode
- Allow non cmd test to build with different versions 
- `scripts/test-sanity.sh` and `scripts/test-wp-cs.sh` work with FIPS and without 
- `scripts/test-openssl.sh` is broken with FIPS, master, and v5.8.0-stable with openssl-3.5.0 no matter what combo. Already disabled in jenkins though. 
- Unsets WPFF when doing `init_wolfprov` so we dont fail unit test for testing

### DIFF
--- a/scripts/cmd_test/aes-cmd-test.sh
+++ b/scripts/cmd_test/aes-cmd-test.sh
@@ -33,14 +33,8 @@ source "${UTILS_DIR}/utils-openssl.sh"
 source "${UTILS_DIR}/utils-wolfssl.sh"
 source "${UTILS_DIR}/utils-wolfprovider.sh"
 
-# Initialize wolfProvider without WPFF set
-if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
-    unset WOLFPROV_FORCE_FAIL
-    init_wolfprov
-    export WOLFPROV_FORCE_FAIL=1
-else 
-    init_wolfprov
-fi
+# Initialize wolfProvider
+init_wolfprov
 
 # Fail flags
 FAIL=0
--- a/scripts/cmd_test/ecc-cmd-test.sh
+++ b/scripts/cmd_test/ecc-cmd-test.sh
@@ -33,14 +33,8 @@ source "${UTILS_DIR}/utils-openssl.sh"
 source "${UTILS_DIR}/utils-wolfssl.sh"
 source "${UTILS_DIR}/utils-wolfprovider.sh"
 
-# Initialize wolfProvider without WPFF set
-if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
-    unset WOLFPROV_FORCE_FAIL
-    init_wolfprov
-    export WOLFPROV_FORCE_FAIL=1
-else 
-    init_wolfprov
-fi
+# Initialize wolfProvider
+init_wolfprov
 
 # Fail flags
 FAIL=0
--- a/scripts/cmd_test/hash-cmd-test.sh
+++ b/scripts/cmd_test/hash-cmd-test.sh
@@ -33,14 +33,8 @@ source "${UTILS_DIR}/utils-openssl.sh"
 source "${UTILS_DIR}/utils-wolfssl.sh"
 source "${UTILS_DIR}/utils-wolfprovider.sh"
 
-# Initialize wolfProvider without WPFF set
-if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
-    unset WOLFPROV_FORCE_FAIL
-    init_wolfprov
-    export WOLFPROV_FORCE_FAIL=1
-else 
-    init_wolfprov
-fi
+# Initialize wolfProvider
+init_wolfprov
 
 # Fail flags
 FAIL=0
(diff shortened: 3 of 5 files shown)


scripts/cmd_test/aes-cmd-test.sh
```diff
@@ -40,13 +40,19 @@ init_wolfprov
...
-    FORCE_FAIL=1
-    echo -e "\nForce fail mode enabled for AES tests"
-fi
+# Check for parameters (order doesn't matter)
+for param in "$1" "$2"; do
+    if [ "$param" = "WOLFPROV_FORCE_FAIL=1" ]; then
+        export WOLFPROV_FORCE_FAIL=1
+        FORCE_FAIL=1
+        echo -e "\nForce fail mode enabled for AES tests"
+    elif [ "$param" = "WOLFPROV_FIPS=1" ]; then
```
padelsbach, line 51: There would be several FIPS variables after this. Some are parameters while others are environment variables.

Would it make sense to use the existing environment variable `WOLFSSL_ISFIPS` like the build script does?

scripts/test-openssl.sh
```diff
@@ -17,13 +17,13 @@
...
-
-# Execute this script from: wolfProvider
-#set -e
+# You can set WOLFPROV_FIPS=1 to run the tests in FIPS mode. You can also set 
+# env variables before running to build with like WOLFSSL_TAG and OPENSSL_TAG etc.
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 NUMCPU=${NUMCPU:-8}
 WOLFPROV_DEBUG=${WOLFPROV_DEBUG:-0}
+WOLFPROV_FIPS=${WOLFPROV_FIPS:-0}
```
padelsbach, line 26: Can we use the same naming convention as `utils-wolfssl.sh` which has `WOLFSSL_ISFIPS`

scripts/test-openssl.sh
```diff
@@ -264,6 +289,11 @@ if [ -z $NUMCPU ]; then
     fi
 fi
 
+# Set WOLFSSL_ISFIPS if FIPS mode is requested
+if [ "$WOLFPROV_FIPS" = "1" ]; then
+    export WOLFSSL_ISFIPS=1
```
padelsbach, line 294: We could simply require the caller to set this

scripts/cmd_test/aes-cmd-test.sh
aidangarske: I like the idea of using `WOLFSSL_ISFIPS`. I had only made it this way because that's how force fail does it but it would be nice to jsut use the env variables. Ill try to use that instead that way we dont have extra special param names and all that.

scripts/cmd_test/aes-cmd-test.sh
```diff
@@ -33,19 +33,29 @@ source "${UTILS_DIR}/utils-openssl.sh"
 source "${UTILS_DIR}/utils-wolfssl.sh"
 source "${UTILS_DIR}/utils-wolfprovider.sh"
 
-# Initialize the environment
-init_wolfprov
+# Initialize wolfProvider without WPFF set
```
padelsbach, line 36: Can you elaborate on why the force fail flag needs to be unset for the `init_wolfprov` call?

scripts/cmd_test/aes-cmd-test.sh
```diff
@@ -33,19 +33,29 @@ source "${UTILS_DIR}/utils-openssl.sh"
...
-    export WOLFPROV_FORCE_FAIL=1
+# Get the force fail parameter
+if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
+    echo "Force fail mode enabled for AES tests"
     FORCE_FAIL=1
-    echo -e "\nForce fail mode enabled for AES tests"
+fi
+if [ "${WOLFSSL_ISFIPS}" = "1" ]; then
+    echo "FIPS mode enabled for AES tests"
+    FIPS=1
```
padelsbach, line 58: Why do we need this duplicate variable?

scripts/cmd_test/aes-cmd-test.sh
```diff
@@ -90,11 +105,14 @@ for key_size in "${KEY_SIZES[@]}"; do
         echo -e "\n=== Testing AES-${key_size}-${mode} ==="
         
         # Generate random key and IV
-        key=$($OPENSSL_BIN rand -hex $((key_size/8)))
+        key=$($OPENSSL_BIN rand -hex $((key_size/8)) 2>/dev/null | tail -n 1 | tr -d '\n')
```
padelsbach, line 108: Just curious, was there an issue with the output of the old command?

padelsbach (commented):
Nice improvement, but a couple concerns

scripts/cmd_test/aes-cmd-test.sh
aidangarske: Yes, If you set the force fail variable before you build the unit test will fail and cause the cmd test to fail. unseting force fail for the build makes sure this doesn't happen.

scripts/cmd_test/aes-cmd-test.sh
aidangarske: We dont ill change that

PR #192, opened 2025-06-25
test github actions for xmlsec

BridgerVoss:
On branch master, had to disable some test that use RSA 512 bit key sizes.
On branch xmlsec-1_2_37, had to disable some test that uses RSA 512 bit key sizes and some tests that use RC2 algorithms.
Otherwise everything works as expected with wolfProvider fixes.

### DIFF
--- a/.github/workflows/xmlsec.yml
+++ b/.github/workflows/xmlsec.yml
@@ -1,0 +1,108 @@
+name: xmlsec Tests
+
+# START OF COMMON SECTION
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+# END OF COMMON SECTION
+
+jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
+      wolfssl_ref: ${{ matrix.wolfssl_ref }}
+      openssl_ref: ${{ matrix.openssl_ref }}
+    strategy:
+      matrix:
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
+
+  test_xmlsec:
+    runs-on: ubuntu-22.04
+    needs: build_wolfprovider
+    # This should be a safe limit for the tests to run.
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        xmlsec_ref: [ 'master', 'xmlsec-1_2_37' ]
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
+        force_fail: [ 'WOLFPROV_FORCE_FAIL=1', '' ]
+        exclude:
+          - xmlsec_ref: 'master'
+            force_fail: 'WOLFPROV_FORCE_FAIL=1'
+    steps:
+      # Checkout the source so we can run the check-workflow-result script
+      - name: Checkout wolfProvider
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Retrieving wolfProvider from cache
+        uses: actions/cache/restore@v4
+        id: wolfprov-cache-restore
+        with:
+          path: |
+            wolfssl-install
+            wolfprov-install
+            openssl-install/lib64
+            openssl-install/include
+            openssl-install/bin
+          key: wolfprov-${{ matrix.wolfssl_ref }}-${{ matrix.openssl_ref }}-${{ github.sha }}
+          fail-on-cache-miss: true
+
+      - name: Install xmlsec dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y automake autoconf libtool libtool-bin \
+              libltdl-dev libltdl7 libxml2-dev
+
+      - name: Checkout OSP
+        uses: actions/checkout@v4
+        with:
+          repository: wolfSSL/osp
+          path: osp
+          fetch-depth: 1
+
+      - name: Download xmlsec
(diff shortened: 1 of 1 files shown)


.github/workflows/xmlsec.yml
```diff
@@ -0,0 +1,122 @@
...
+          repository: lsh123/xmlsec
+          ref: ${{ matrix.xmlsec_ref }}
+          path: xmlsec
+
+      - name: Build xmlsec
+        working-directory: xmlsec
+        env:
+          XMLSEC_REF: ${{ matrix.xmlsec_ref }}
+        run: |
+          ./autogen.sh --disable-openssl3-engines --disable-dsa --without-nss --without-gnutls --without-gcrypt --disable-xmldsig --disable-crypto-dl --disable-apps-crypto-dl --disable-concatkdf --disable-pbkdf2 --disable-tmpl-tests
```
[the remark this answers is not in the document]
ColtonWilley: I reviewed these options and they do look OK to me. Many are disabling alternate crypto implementations, odd options related to dynamic linking. The only options that do look questionable to me is "--disable-concatkdf --disable-pbkdf2", but those only apply for master, not the version targeted for IGEL. Looks good to me as is.

.github/workflows/xmlsec.yml
```diff
@@ -0,0 +1,127 @@
...
+
+          XMLSEC_REF=${{ matrix.xmlsec_ref }}
+          if [ "$XMLSEC_REF" = "master" ]; then
+            echo "Running commands for master"
+            # Disables test using RSA512
+            sed -i '416,432s/^/#/' tests/testEnc.sh
+          elif [ "$XMLSEC_REF" = "xmlsec-1_2_37" ]; then
+            echo "Running commands for xmlsec-1_2_37"
+            # Disables tests using RSA512
+            sed -i '33,47s/^/#/' tests/testEnc.sh
```
ColtonWilley, line 104: We have most of our other test stripping contained within an OSP patch. This would be a bit inconsistent, @padelsbach thoughts?

padelsbach: True. If the change is specific to wolfProvider test cases, I think it's ok to live here. If it's changing the OSP to work with provider outside of testing or with wolfSSL or FIPS, etc, I would say it should be in OSP.

But I understand if it's better to put _all_ patches in OSP.

ColtonWilley: The inconsistency is killing me. I would prefer all OSP modifications in one place. @BridgerVoss make these changes into an osp patch and instead apply it here.

.github/workflows/xmlsec.yml
```diff
@@ -0,0 +1,128 @@
...
+        working-directory: xmlsec
+        env:
+          XMLSEC_REF: ${{ matrix.xmlsec_ref }}
+        run: |
+          ./autogen.sh --disable-openssl3-engines --disable-dsa --without-nss \
+              --without-gnutls --without-gcrypt --disable-xmldsig \
+              --disable-crypto-dl --disable-apps-crypto-dl \
+              --disable-concatkdf --disable-tmpl-tests
+          make
+          cat <<EOF > tests/openssl3.cnf
```
padelsbach, line 84: This looks mostly identical to the existing `provider.conf` file, and does not comprehend FIPS. Is there a reason for not using the one in our repo?

BridgerVoss: I was going to copy the one from our repo into this one because I think that is better in case of code changes to provider.conf, but it essentially hard loads this conf file in a bunch of places so I think this way is a lot easier than patching a bunch of files.

ColtonWilley: Agreed, would prefer to copy over existing provider.conf instead.

ColtonWilley: Having it still end up in openssl3.conf is fine, but copy over the existing provider.conf from wolfProvider instead. We _want_ any changes in provider.conf to be immediately picked up by the test.

.github/workflows/xmlsec.yml
```diff
@@ -0,0 +1,108 @@
...
+          ref: ${{ matrix.xmlsec_ref }}
+          path: xmlsec
+          fetch-depth: 1
+
+      - name: Build xmlsec
+        working-directory: xmlsec
+        env:
+          XMLSEC_REF: ${{ matrix.xmlsec_ref }}
+        run: |
+          ./autogen.sh --disable-openssl3-engines --disable-dsa --without-nss \
```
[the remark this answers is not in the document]
BridgerVoss: `autogen.sh` runs configure in this repo and make is right after this

.github/workflows/xmlsec.yml
```diff
@@ -0,0 +1,108 @@
...
+              --without-gnutls --without-gcrypt --disable-xmldsig \
+              --disable-crypto-dl --disable-apps-crypto-dl \
+              --disable-concatkdf --disable-tmpl-tests
+          make
+          cp $GITHUB_WORKSPACE/provider.conf tests/openssl3.cnf
+          patch -p1 < $GITHUB_WORKSPACE/osp/wolfProvider/xmlsec/xmlsec-${{ matrix.xmlsec_ref }}-wolfprov.patch
```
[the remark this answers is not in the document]
BridgerVoss: The patch is for a bash script only but I'll move it before build related stuff if you want. I don't think i can apply it in the `checkout@v4` step because it won't apply the patch unless I am in the xmlsec repo?

.github/workflows/xmlsec.yml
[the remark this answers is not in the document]
padelsbach: Ah ok

.github/workflows/xmlsec.yml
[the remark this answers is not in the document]
padelsbach: Just a best practice since other yamls modify the build scripts. 

Guessing it is possible to clone the xmlsec repo first, then clone osp. Then `cd xmlsec` and apply the patch.

padelsbach (approved):

PR #175, opened 2025-06-16
libhashkit2 yml workflow

BridgerVoss:
Added wolfProvider tests for libhashkit2 which is a part of the bigger repository, libmemcached.
- libhashkit is passing all tests now as expected with updated wolfProvider

PR #193, opened 2025-06-26
Fix IV handling for consecutive AES CBC calls

padelsbach:
This change fixes an issue encountered when integrating libhashkit and libmemcached in this PR: https://github.com/wolfSSL/wolfProvider/pull/175

### DIFF
--- a/scripts/env-setup
+++ b/scripts/env-setup
@@ -13,6 +13,16 @@ if [ $is_sourced -eq 0 ]; then
     exit 1
 fi
 
+
+if [ -n "$BASH_SOURCE" ]; then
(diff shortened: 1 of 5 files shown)


scripts/env-setup
```diff
@@ -13,6 +13,16 @@ if [ $is_sourced -eq 0 ]; then
     exit 1
 fi
 
+
+if [ -n "$BASH_SOURCE" ]; then
```
padelsbach, line 17: This is an unrelated change, but helps with the zsh/bash use cases

SparkiDev (dismissed):

douzzer:
@padelsbach this has a conflict now -- probably caused by my merging #189.

PR #201, opened 2025-07-02
Made x11vnc tests and fixed rsa certificate display

sebastian-carpenter:
Made a testsuit for x11vnc in lieu of a make check.
Private and public key display was not handled. Paul gave me some code that addressed the bug then I reworked it to match the output from OSSL and look pretty.

### DIFF
--- a/.github/workflows/x11vnc.yml
+++ b/.github/workflows/x11vnc.yml
@@ -89,6 +89,7 @@ jobs:
       - name: Run x11vnc tests
         run: |
           source $GITHUB_WORKSPACE/scripts/env-setup
+          export ${{ matrix.force_fail }}
           export WOLFPROV_FORCE_FAIL_STR="${{ matrix.force_fail }}"
 
           $GITHUB_WORKSPACE/.github/scripts/x11vnc/test_x11vnc.sh


src/wp_rsa_kmgmt.c
```diff
@@ -3840,5 +3840,320 @@ const OSSL_DISPATCH wp_rsapss_pki_pem_encoder_functions[] = {
...
+    (void)pwCb;
+    (void)pwCbArg;
+
+    if (out == NULL) {
+        ok = 0;
+    }
+
+    /* Calculate total size needed for text output */
+    if (ok) {
+        textLen = 128;
```
padelsbach, line 4017: Can you add a comment about the calculation of `textLen` and why 128+bits*2 is needed?

scripts/env-setup
```diff
@@ -13,7 +13,15 @@ if [ $is_sourced -eq 0 ]; then
     exit 1
 fi
 
-SCRIPT_DIR="$(cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)"
+if [ -n "$BASH_SOURCE" ]; then
```
padelsbach, line 16: Looks like there are conflicts in this file. I suspect it should match what is on master.

src/wp_rsa_kmgmt.c
```diff
@@ -3840,5 +3840,320 @@ const OSSL_DISPATCH wp_rsapss_pki_pem_encoder_functions[] = {
...
+                ok = 0;
+            }
+            dPos += printAmt;
+        }
+
+        OPENSSL_free(binData);
+        binData = NULL;
+    }
+
+    *pos = dPos;
```
padelsbach, line 3975: Nit: I usually like the practice of not modifying the callers variables when a failure is encountered. Can we wrap line 3975 with an `if (ok)`

src/wp_rsa_kmgmt.c
```diff
@@ -3840,5 +3840,320 @@ const OSSL_DISPATCH wp_rsapss_pki_pem_encoder_functions[] = {
...
+        ok = 0;
+    }
+    else if (mp_to_unsigned_bin(num, binData) != MP_OKAY) {
+        ok = 0;
+        OPENSSL_free(binData);
+        binData = NULL;
+    }
+    else {
+        if ((printAmt = XSNPRINTF(textData + dPos, textLen - dPos,
+                "%s:\n    ", label)) <= 0){
```
padelsbach, line 3928: Nit: Wolf seems to use the convention of having a space between the condition and the brace. Can you please update that throughout the file?

.github/workflows/x11vnc.yml
```diff
@@ -0,0 +1,428 @@
...
+          else
+              echo "[ FAILED ] -https with an https request"
+              X11VNC_TEST_FAIL=1
+          fi
+
+          killall x11vnc > /dev/null 2> /dev/null
+          killall Xvfb > /dev/null 2> /dev/null
+          cat server.log >> x11vnc_test.log
+
+          if [[ $WOLFPROV_FORCE_FAIL -eq 1 ]]; then
```
padelsbach, line 259: Instead of this, can we use `check-workflow-result.sh` like the other yamls do?

src/wp_rsa_kmgmt.c
sebastian-carpenter: added some comments describing it

src/wp_rsa_kmgmt.c
sebastian-carpenter: wrapped it with an ok check

src/wp_rsa_kmgmt.c
sebastian-carpenter: updated braces to not be directly adjacent. Also changed all else and else ifs to be on new lines

.github/workflows/x11vnc.yml
```diff
@@ -0,0 +1,428 @@
...
+          perl -pi -e 's/-des3/-aes256/' src/ssltools.h
+
+          autoreconf -vfi
+          ./configure --with-ssl="$GITHUB_WORKSPACE/openssl-install/lib64"
+          make -j $(nproc)
+          sudo make install
+
+      - name: Create x11vnc main test
+        run: |
+          cat > test_x11vnc.sh << 'TEST_X11VNC_EOF'
```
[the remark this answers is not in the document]
sebastian-carpenter: main tests and expect scripts have been broken into .github/scripts/x11vnc

.github/workflows/x11vnc.yml
sebastian-carpenter: updated, checking if my solution works as expected in the action

sebastian-carpenter: action passed

padelsbach (approved):

PR #202, opened 2025-07-03
Add Test for systemd

ejohnstown:
1. Update the check workflow result to add a systemd section.
2. Add an action script for systemd.

### DIFF
--- a/.github/workflows/systemd.yml
+++ b/.github/workflows/systemd.yml
@@ -1,0 +1,97 @@
+name: systemd Tests
+
+# START OF COMMON SECTION
+on:
+  push:
+    branches: ['master', 'main', 'release/**']
+  pull_request:
+    branches: ['*']
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+# END OF COMMON SECTION
+
+jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
+      wolfssl_ref: ${{ matrix.wolfssl_ref }}
+      openssl_ref: ${{ matrix.openssl_ref }}
+    strategy:
+      matrix:
+        wolfssl_ref: ['v5.8.0-stable', 'master']
+        openssl_ref: ['openssl-3.5.0']
+
+  test_systemd:
+    runs-on: ubuntu-22.04
+    needs: build_wolfprovider
+    # This should be a safe limit for the tests to run.
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        systemd_ref: ['v254']
+        wolfssl_ref: ['v5.8.0-stable', 'master']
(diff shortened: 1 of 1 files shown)


.github/scripts/check-workflow-result.sh
```diff
@@ -294,6 +294,22 @@ if [ "$WOLFPROV_FORCE_FAIL" = "WOLFPROV_FORCE_FAIL=1" ]; then
             echo "Error: $IPERF_TEST_LOG not found"
             exit 1
         fi
+    # ----- SYSTEMD -----
```
aidangarske, line 297: Perhaps handle this in the .yml file rather than here and utilize the catch at end of this script since we are not checking for anything specific.

.github/workflows/systemd.yml
```diff
@@ -0,0 +1,93 @@
...
+        systemd_ref: ['v254']
+        wolfssl_ref: ['v5.8.0-stable', 'master']
+        openssl_ref: ['openssl-3.5.0']
+        force_fail: ['WOLFPROV_FORCE_FAIL=1', '']
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential meson ninja-build \
+            libssl-dev libmount-dev gperf python3-pytest libuv1-dev \
```
aidangarske, line 43: Lets not use libssl-dev

aidangarske (changes requested):

ejohnstown:
Jenkins retest this please

.github/workflows/systemd.yml
```diff
@@ -0,0 +1,109 @@
...
+                      resolvectl systemd-resolved test-cryptolib \
+                      test-dns-packet test-dnssec test-resolve-tables \
+                      test-resolved-etc-hosts test-resolved-packet \
+                      test-resolved-stream"
+          if [ -n ${{ matrix.force_fail }} ]; then
+            export ${{ matrix.force_fail }}
+          fi
+          meson test -C build $TEST_CASES
+          TEST_RESULT=$?
+          if [ -z ${{ matrix.force_fail }} ]; then
```
aidangarske, line 97: This logic is fine but ideally we would utilize the logic in place in the `check-worklfow-results.sh`. Consider doing something like `tnftp.yml` or `cjose,yml`. Just grep for the failure and return 1 or 0 and feed that into test result.

.github/workflows/systemd.yml
```diff
@@ -0,0 +1,97 @@
...
+        working-directory: systemd
+        run: |
+          source $GITHUB_WORKSPACE/scripts/env-setup
+          meson setup -Dnobody-group=nogroup build
+          ninja -C build
+
+      - name: Run systemd tests
+        working-directory: systemd
+        run: |
+          set +e
```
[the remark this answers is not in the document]
ejohnstown: "set +e" is very necessary. When the `meson test` case fails because we are forcing a failure, the run fails and exits before running the check script.

PR #207, opened 2025-07-03
Fix ECC public key validation

ColtonWilley:
Fix an issue with ECC public key validation. When asked to validate the public key, if a private key is present we also check the public against the private. This does not work for a KRB5 use case where there is a private value leftover from the original key the key being used was duped from. In this case the public and private keys do not match up, which is fine because we only need to check the private. Needed a bit of logic to trick wolfCrypt into doing ECC public key only validation.

### DIFF
--- a/src/wp_ecc_kmgmt.c
+++ b/src/wp_ecc_kmgmt.c
@@ -1995,6 +1995,28 @@ static int wp_ecc_decode_params(wp_Ecc* ecc, unsigned char* data, word32 len)
     return ok;
 }
 
+static int wp_ecc_decode_x963_pub(wp_Ecc* ecc, unsigned char* data, word32 len)
+{
+    int ok = 1;
+    int rc;
+
+    rc = wc_ecc_import_x963((const byte *)data, len, &ecc->key);
+    if (rc != 0) {
+        ok = 0;
+    }
+    if (ok) {
+        ecc->curveId = ecc->key.dp->id;
+        ecc->hasPub = 1;
+        /* Needs curveId set. */
+        if (!wp_ecc_set_bits(ecc)) {
+            ok = 0;
+        }
+    }
+
+    WOLFPROV_LEAVE(WP_LOG_PK, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
+    return ok;
+}
+
 /**
  * Decode the SubjectPublicInfo DER encoded ECC key into the ECC key object.
  *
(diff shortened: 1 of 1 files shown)


src/wp_ecc_kmgmt.c
```diff
@@ -1004,8 +1004,17 @@ static int wp_ecc_validate(const wp_Ecc* ecc, int selection, int checkType)
...
     #endif
         {
-            privDone = 1;
+            /* We may have a private key that does not correspond to the public
+             * key, which is fine as we are only being asked to check the
+             * public at this time. wc_ecc_check_key() always checks both if it
+             * has a private key, so we will fool it into only checking public
+             * key by manually setting the type */
+            origType = ecc->key.type;
+            if (ecc->key.type != ECC_PUBLICKEY) {
```
SparkiDev, line 1013: Just assign anyway.
If threading is going to be an issue then copy out public key into new object. (I assume not but had to say it.)

SparkiDev (changes requested):

src/wp_ecc_kmgmt.c
```diff
@@ -980,8 +980,8 @@ static int wp_ecc_validate_public_key_quick(const wp_Ecc* ecc)
 static int wp_ecc_validate(const wp_Ecc* ecc, int selection, int checkType)
 {
     int ok = 1;
-    int privDone = 0;
     int rc;
+    int origType;
 
     /* Only named curves supported. */
     if (((selection & OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS) != 0) &&
```
padelsbach, line 987: I know it's not part of this change, but this function does a bunch of work even when `ok == 0`

src/wp_ecc_kmgmt.c
padelsbach, line 1013: Why is the cast needed on line 1014 and 1017, but not on line 1013? Maybe I'm missing something

src/wp_ecc_kmgmt.c
```diff
@@ -1004,8 +1004,17 @@ static int wp_ecc_validate(const wp_Ecc* ecc, int selection, int checkType)
...
         {
-            privDone = 1;
+            /* We may have a private key that does not correspond to the public
+             * key, which is fine as we are only being asked to check the
+             * public at this time. wc_ecc_check_key() always checks both if it
+             * has a private key, so we will fool it into only checking public
+             * key by manually setting the type */
+            origType = ecc->key.type;
+            if (ecc->key.type != ECC_PUBLICKEY) {
+                ((wp_Ecc*)ecc)->key.type = ECC_PUBLICKEY;
```
padelsbach, line 1014: Can we just set `type = ECC_PUBLICKEY` or does the check add something?

padelsbach (approved):
Looks good, nice comments. Couple questions/concerns

SparkiDev (dismissed):

aidangarske (dismissed):

PR #208, opened 2025-07-07
BIO_printf refactor for RSA certificate display

sebastian-carpenter:
updated RSA key display to use BIO_printf instead of XSNPRINTF

### DIFF
--- a/src/wp_rsa_kmgmt.c
+++ b/src/wp_rsa_kmgmt.c
@@ -4068,21 +4068,17 @@ static int wp_rsa_text_enc_does_selection(WOLFPROV_CTX* provCtx, int selection)
 /**
  * Encode a portion of the RSA key as hexadecimal.
  *
+ * @param [out]     out        BIO to write data out onto.
  * @param [in]      num        An mp_int to convert to hexadecimal.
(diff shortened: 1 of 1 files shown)


src/wp_rsa_kmgmt.c
```diff
@@ -4026,52 +4005,27 @@ static int wp_rsa_encode_text(wp_RsaEncDecCtx* ctx, OSSL_CORE_BIO* cBio,
...
-            ok = 0;
-        }
-    }
-
     if (ok) {
         /* OSSL uses nested macros to determine the number of primes, not sure
          * when there wouldn't be two primes */
-        if (hasPriv && (printAmt = XSNPRINTF(textData + pos, textLen - pos,
-                "Private-Key: (%d bit, 2 primes)\n", key->bits)) <= 0) {
+        if (hasPriv && BIO_printf(out, "Private-Key: (%d bit, 2 primes)\n",
```
padelsbach, line 4011: The length checks in the old code are crucial for ensuring we don't overflow the buffer. `BIO_snprintf` or similar is a better replacement.

ColtonWilley (approved):

PR #210, opened 2025-07-07
fixed timeout errors for x11vnc workflow

sebastian-carpenter:
Too small of a timeout causes sslgencert_client/server or sslgenca to fail periodically. Increased timeouts across the board to help mitigate this.

Created a loop with the offending tests and it took me about half an hour to get an error.
After fixing it I ran that same loop again for over 3 hours and have not gotten an error.

Presumably, it is fixed.


padelsbach (approved):
Looks good, thanks for the quick fix

PR #212, opened 2025-07-09
Add ref for pam-pkcs11 test and fix WPFF

aidangarske:


.github/scripts/pam-pkcs11-test.sh
```diff
@@ -63,6 +63,10 @@ fi
 
 echo "[*] Configuring pam_pkcs11..."
 
+# Temporarily unset WOLFPROV_FORCE_FAIL so we can test the failure case
```
padelsbach, line 66: Comment could be more clear about why WPFF needs to be unset here

padelsbach (approved):
Nice catch! Thanks for fixing

ColtonWilley (dismissed):

ColtonWilley (approved):

PR #220, opened 2025-07-17
Modify core libctx handling to instead create a new child libctx

ColtonWilley:
There are certain initialization flows that yield an uninitialized libctx pointer, which is a problem when we go to fetch for something like KDF later. Instead attempt to create a new child libctx, which is associated with the parent ctx to be instantiated.


ColtonWilley:
Retest this please.

SparkiDev (approved):

PR #225, opened 2025-07-18
Fix tag logic in build script

aidangarske:
# Description 

Our nightly's are failing [here](https://cloud.wolfssl-test.com/jenkins/job/wolfProvider/job/nightly-scripts-test/399/pipeline-console/) with wolfssl master because the current tag is `v5.8.2-stable` while it is expecting master. This changes the logic so it prioritizes `git branch --show-current` for master and `git describe --tags` for `*-stable || v*`. It will fall back if none of those work to the opposite just like it did before. 
- Now it can also take commit hashes like `b077c81eb635392e694ccedbab8b644297ec0285` or `b077c81` when building by setting `WOLFSSL_TAG` or `OPENSSL_TAG` with the commit.

### DIFF
--- a/scripts/utils-openssl.sh
+++ b/scripts/utils-openssl.sh
@@ -39,16 +39,7 @@ USE_CUR_TAG=${USE_CUR_TAG:-0}
 
 clone_openssl() {
     if [ -d ${OPENSSL_SOURCE_DIR} ] && [ "$USE_CUR_TAG" != "1" ]; then
-        if [[ "${OPENSSL_TAG}" == openssl-* ]]; then
-            OPENSSL_TAG_CUR=$(cd ${OPENSSL_SOURCE_DIR} && (git describe --tags 2>/dev/null || git branch --show-current 2>/dev/null))
-        else
-            OPENSSL_TAG_CUR=$(cd ${OPENSSL_SOURCE_DIR} && (git branch --show-current 2>/dev/null || git describe --tags 2>/dev/null))
-        fi
-        if [ "${OPENSSL_TAG_CUR}" != "${OPENSSL_TAG}" ]; then # force a rebuild
-            printf "Version inconsistency. Please fix ${OPENSSL_SOURCE_DIR} (expected: ${OPENSSL_TAG}, got: ${OPENSSL_TAG_CUR})\n"
-            do_cleanup
-            exit 1
-        fi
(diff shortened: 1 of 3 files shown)


scripts/utils-openssl.sh
```diff
@@ -39,7 +39,11 @@ USE_CUR_TAG=${USE_CUR_TAG:-0}
 
 clone_openssl() {
     if [ -d ${OPENSSL_SOURCE_DIR} ] && [ "$USE_CUR_TAG" != "1" ]; then
-        OPENSSL_TAG_CUR=$(cd ${OPENSSL_SOURCE_DIR} && (git describe --tags 2>/dev/null || git branch --show-current))
+        if [[ "${OPENSSL_TAG}" == openssl-* ]]; then
```
padelsbach, line 42: I'm thinking we could make this a bit more general-purpose to see if `OPENSSL_TAG` matches a tag, branch or commit hash, and perhaps live in utils-general.sh. What do you think?

AI helped me generate this: 
```
check_git_match() {
    local FOO="$1"

    local TAG BRANCH COMMIT_LONG COMMIT_SHORT
    TAG=$(git describe --tags --exact-match 2>/dev/null || true)
    BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null || true)
    COMMIT_LONG=$(git rev-parse HEAD)
    COMMIT_SHORT=$(git rev-parse --short HEAD)

    if [[ "$FOO" == "$TAG" ]]; then
        echo "match: tag ($TAG)"
        return 0
    elif [[ "$FOO" == "$BRANCH" ]]; then
        echo "match: branch ($BRANCH)"
        return 0
    elif [[ "$FOO" == "$COMMIT_LONG" ]]; then
        echo "match: commit hash (long $COMMIT_LONG)"
        return 0
    elif [[ "$FOO" == "$COMMIT_SHORT" ]]; then
        echo "match: commit hash (short $COMMIT_SHORT)"
        return 0
    else
        echo "no match"
        return 1
    fi
}
```

aidangarske: Ooh yes, I like that more. Ill make a function in general =

padelsbach: Nice!

padelsbach (dismissed):

padelsbach (approved):
Nice work, thanks

PR #229, opened 2025-07-22
added sscep tests

sebastian-carpenter:
This tests the bare minimum of sscep. A better scep server is needed for further testing or the current scep server needs to be patched to not use des-cbc.

### DIFF
--- a/.github/scripts/test_sscep.sh
+++ b/.github/scripts/test_sscep.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+# test_sscep.sh
+#
+# Copyright (C) 2006-2025 wolfSSL Inc.
+#
+# This file is part of wolfProvider.
+#
+# wolfProvider is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# wolfProvider is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+TEST_SSCEP_FAIL=0
+
+cleanup(){
+    [ -f ca.crt     ] && rm -f  ca.crt
+    [ -d ca-dir     ] && rm -rf ca-dir
+}
+
+killall scepserver &> /dev/null
+cleanup
+
+# begin by setting up and starting the scep server
+OPENSSL_CONF="" OPENSSL_MODULES="" scepserver ca -depot ca-dir -init
+OPENSSL_CONF="" OPENSSL_MODULES="" scepserver -depot ca-dir -port 8080 -debug &
+
+sleep 1
+
+# now test sscep
+
+# getca
+sscep getca -u "http://localhost:8080/scep" -c ca.crt -v -d
+
+if [ $? -eq 0 ] && [ -f ca.crt ] \
+    && diff -y ca.crt ca-dir/ca.pem
+then
+    echo "[ PASSED ] getca"
+else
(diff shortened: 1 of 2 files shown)


.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
+TEST_SSCEP_FAIL=0
+
+cleanup(){
+    [ -f ca.crt     ] && rm -f  ca.crt
+    [ -d ca-dir     ] && rm -rf ca-dir
+    [ -f client.log ] && rm -f  client.log
+    [ -f server.log ] && rm -f  server.log
+}
+
+killall scepserver >> /dev/null 2>> /dev/null
```
padelsbach, line 10: Probably more concise to write: `killall scepserver &> /dev/null`

.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
+TEST_SSCEP_FAIL=0
```
padelsbach, line 1: Please add the copyright header and a description of this file

.github/scripts/test_sscep.sh
padelsbach: And a few other places in this file could benefit from the `&>` operator instead of the double redirect

.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
...
+# could not get certificate chaining to work. Not sure if it's the servers fault
+#   or mine.
+
+# enroll
+# first generate ca request (sscep has a script for this)
+timeout 10 ./mkrequest -ip 1.2.3.4
+
+if [ $? -eq 0 ]; then
+    # then enroll -> sscep WILL fail this.
+    # scepserver uses des-cbc (which is not supported) when sending a cert back, so
```
padelsbach, line 45: How hard would it be to patch scepserver to use aes-cbc instead?

.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
...
+
+killall scepserver >> /dev/null 2>> /dev/null
+
+echo -e "\n\n\nclient log:\n\n\n" | cat server.log - client.log > test_sscep.log
+cleanup
+
+$GITHUB_WORKSPACE/.github/scripts/check-workflow-result.sh $TEST_SSCEP_FAIL "$WOLFPROV_FORCE_FAIL_STR" sscep
+TEST_SSCEP_FAIL=$?
+
+if [[ $TEST_SSCEP_FAIL -eq 1 ]]; then cat test_sscep.log; fi
```
padelsbach, line 70: Seems like the log could be quite long. Consider just printing a message to see the log file for details.

.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
...
+    TEST_SSCEP_FAIL=1
+fi
+
+killall scepserver >> /dev/null 2>> /dev/null
+
+echo -e "\n\n\nclient log:\n\n\n" | cat server.log - client.log > test_sscep.log
+cleanup
+
+$GITHUB_WORKSPACE/.github/scripts/check-workflow-result.sh $TEST_SSCEP_FAIL "$WOLFPROV_FORCE_FAIL_STR" sscep
+TEST_SSCEP_FAIL=$?
```
padelsbach, line 68: This updates `TEST_SSCEP_FAIL` to be the result of the `check-workflow` script. Why change the value of `TEST_SSCEP_FAIL` here?

.github/scripts/test_sscep.sh
```diff
@@ -0,0 +1,72 @@
...
+        TEST_SSCEP_FAIL=1
+    fi
+else
+    echo "[ FAILED ] enroll"
+    TEST_SSCEP_FAIL=1
+fi
+
+killall scepserver >> /dev/null 2>> /dev/null
+
+echo -e "\n\n\nclient log:\n\n\n" | cat server.log - client.log > test_sscep.log
```
padelsbach, line 64: I'd vote to keep the logs separate, if possible

.github/scripts/test_sscep.sh
padelsbach: Ok, thanks for checking. Let's leave it for now.

padelsbach (approved):
Looks good, nice work. Can merge when tests complete.

PR #232, opened 2025-07-24
rsync workflow

JeremiahM37:
Created a basic test for sha 256 and sha512, which are the only fips approved algorithms that rsync uses. It has its own implementations of MD4 and MD5 used for file checksums and checksum transfers, which I commented how to disable, but have not removed them for now.


.github/workflows/rsync.yml
```diff
@@ -0,0 +1,156 @@
...
+        uses: actions/checkout@v4
+        with:
+          repository: RsyncProject/rsync
+          path: rsync_repo
+          ref: ${{ matrix.rsync_ref }}
+          fetch-depth: 1
+
+      - name: Create SHA test script
+        run: |
+          cat > rsync_repo/testsuite/sha-test.test << 'EOF'
```
padelsbach, line 72: Can you move this to it's own file so that we can use it in Jenkins?

.github/workflows/rsync.yml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Build and install rsync
+        working-directory: rsync_repo
+        run: |
+          # Set up the environment for wolfProvider
+          source $GITHUB_WORKSPACE/scripts/env-setup
+          ./configure --disable-xxhash
+          make -j$(nproc)
+          #export RSYNC_CHECKSUM_LIST="none"
+          #This can disable file checksums which currently use rsycs own implementation of MD4 and MD5
```
padelsbach, line 137: `RSYNC_CHECKSUM_LIST=none` seems like something we want so we can test the provider/wolfssl versions. Any reason to not enable it?

.github/workflows/rsync.yml
JeremiahM37: Where should I put that file? Should It go into the osp repo or the wolfprovider/.github/scripts section?

.github/workflows/rsync.yml
padelsbach: I'd say those hash algorithms are pretty well tested, so no need to add a patch just for this. But please add the note about "causes a lot of tests to fail" to the comments.

padelsbach (approved):

PR #233, opened 2025-07-24
Fix opensc osp patch

padelsbach:


.github/workflows/opensc.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           repository: wolfssl/osp
           path: osp
           fetch-depth: 1
-        run: |
+      - run: |
```
aidangarske, line 79: Don't we normally do the run without the - in front of it? How is this different?

PR #234, opened 2025-07-28
Debian package build script

padelsbach:
Add ability to build a Debian package via `build-wolfprovider.sh --debian`, including a github action to test it.

FIPS mode will come next as a follow-up PR.

Note: this PR requires #245 to be merged first.


.github/workflows/debian-package.yml
```diff
@@ -0,0 +1,328 @@
...
+        # Fix any missing dependencies
+        sudo apt-get install -f -y
+        
+        # Verify installation
+        echo "Package Installation Verification:"
+        dpkg -l | grep libwolfprov
+        dpkg -L libwolfprov
+        
+    - name: Test OpenSSL provider functionality
+      run: |
```
aidangarske, line 101: Any reason why we couldn't just call our `./scripts/cmd_tests/do-cmd-test.sh`?

.github/workflows/debian-package.yml
```diff
@@ -0,0 +1,328 @@
...
+        mkdir -p "$RUNNER_TEMP/wolfssl-pkg"
+        cd "$RUNNER_TEMP/wolfssl-pkg"
+        unzip $GITHUB_WORKSPACE/.github/packages/debian-packages-20250731T171211Z-1-001.zip
+        cd debian-packages
+        sudo dpkg-source -x wolfssl_5.8.2-1.dsc
+        cd wolfssl-5.8.2
+        sudo dpkg-buildpackage -b -us -uc
+        sudo dpkg -i ../libwolfssl*.deb
+
+    - name: Build Debian package
```
aidangarske, line 65: Should we use `./scripts/build-wolfprovider.sh --debian` at some point to build the debian package file each time we run the workflow?

.github/workflows/debian-package.yml
```diff
@@ -0,0 +1,328 @@
...
+          sudo mv /tmp/openssl-test/wolfprovider.conf.backup /etc/ssl/openssl.cnf.d/wolfprovider.conf
+          echo "   - Restored wolfprovider.conf to /etc/ssl/openssl.cnf.d/"
+        fi
+        
+        echo "PASS: All provider interoperability tests successful"
+        
+    - name: Uninstall package and verify cleanup
+      run: |
+        # Uninstall the package
+        sudo apt-get remove --purge -y libwolfprov
```
aidangarske, line 264: Question, if we are going to have a `libwolfprov` package like this would we be able to do like a `sudo-apt install libwolfprov` at some point? Or will we always have to build the .deb from a cloned build of wolfProvider?

.github/workflows/debian-package.yml
```diff
@@ -0,0 +1,328 @@
...
+        echo "Package uninstallation and cleanup verification successful"
+
+    - name: Move package artifacts
+      run: |
+        # Move the generated packages to the temp directory
+        mv ../*.deb $RUNNER_TEMP/ || true
+        mv ../*.dsc $RUNNER_TEMP/ || true
+        mv ../*.tar.gz $RUNNER_TEMP/ || true
+        
+    - name: Upload package artifacts
```
aidangarske, line 319: Should we be uploading artifacts all the time? Whats the benefit of doing this even if we pass?

.github/workflows/debian-package.yml
padelsbach: Besides line 73?

.github/workflows/debian-package.yml
padelsbach: I honestly wasn't aware of these until today! Thanks for enlightening me. ;)

The Debian package install currently overrides the default provider to ensure the system is using WP. I'll see if there is a way to bypass that for this test step.

.github/workflows/debian-package.yml
padelsbach: Definitely don't want to clone the source. The `dpkg` on line 90 does effectively the same as `apt`. I'll switch it to `apt` since that is most likely what the user would do.

.github/workflows/debian-package.yml
aidangarske: No not besides that my bad :)

scripts/build-debian.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
aidangarske, line 1: Can we add the wolfprov copyright header here

aidangarske (commented):
I am fine with these changes now

PR #237, opened 2025-07-28
Add kmod CI Workflow

aidangarske:
# Description 

- Add kmod workflow for `v33` 
- WPFF doesn't do anything since we only parse so no need for WPFF in this case
- [ ] needs patch [here](https://github.com/wolfSSL/osp/pull/275)

### DIFF
--- a/.github/workflows/kmod.yml
+++ b/.github/workflows/kmod.yml
@@ -1,0 +1,113 @@
+name: kmod Tests
+
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**' ]
+  pull_request:
+    branches: [ '*' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
+      wolfssl_ref: ${{ matrix.wolfssl_ref }}
+      openssl_ref: ${{ matrix.openssl_ref }}
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/kmod.yml
```diff
@@ -0,0 +1,114 @@
...
+    runs-on: ubuntu-22.04
+    needs: build_wolfprovider
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        kmod_ref: [ 'v33' ]
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
+        # Note: No WOLFPROV_FORCE_FAIL needed - kmod is only a signature parser,
```
padelsbach, line 34: If this is true, then are we actually testing wolfprovider then?

aidangarske: Because WP is used to parse it just doesn't verify. WPFF shouldn't have any effect on the parsing functions that are used so there is no reason to test WPFF. We should still test WP for the parsing functions however since WP does use these.

aidangarske: Yep it is. Honestly the main reason is for IGEL

padelsbach: Any idea why we don't use WPFF in the parsing functions? Should we add it? Maybe @ColtonWilley can chime in here

ColtonWilley: I agree with @padelsbach if there is no crypto usage this should not be a workflow

PR #238, opened 2025-07-29
Use openvpn osp patch

aidangarske:
# Description 

- [x] needs patch [here](https://github.com/wolfSSL/osp/pull/278)


padelsbach (approved):
Looks good, thanks

PR #242, opened 2025-08-01
fixes legacy FIPS wolfProvider support commit

BridgerVoss:
Fixes defined but not used warnings/errors when built with a wolfSSL FIPS that does not include md5 or md5-sha1

### DIFF
--- a/src/wp_digests.c
+++ b/src/wp_digests.c
@@ -220,24 +220,21 @@ const OSSL_DISPATCH name##_functions[] = {                                     \
     { 0,                                NULL                                }  \
 };
 
-#if defined(HAVE_FIPS) && !defined(WP_ALLOW_NON_FIPS)
 #define IMPLEMENT_DIGEST_NULL(name)                                            \
-/** Dispatch table for digest algorithms. */                                   \
+/** Dispatch table for digest_null algorithms. */                              \
+static int name##_digest_null(void) { return 0; }                              \
+static void name##_digest_void(void) {}                                        \
 const OSSL_DISPATCH name##_functions[] = {                                     \
-    { OSSL_FUNC_DIGEST_NEWCTX,          (DFUNC)wp_digest_null               }, \
-    { OSSL_FUNC_DIGEST_INIT,            (DFUNC)wp_digest_null               }, \
-    { OSSL_FUNC_DIGEST_UPDATE,          (DFUNC)wp_digest_null               }, \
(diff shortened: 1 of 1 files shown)


src/wp_digests.c
```diff
@@ -220,7 +220,8 @@ const OSSL_DISPATCH name##_functions[] = {                                     \
     { 0,                                NULL                                }  \
 };
 
-#if defined(HAVE_FIPS) && !defined(WP_ALLOW_NON_FIPS)
+#if defined(HAVE_FIPS) && !defined(WP_ALLOW_NON_FIPS) && (defined(WP_HAVE_MD5) \
```
padelsbach, line 223: This conditional must match the one around line 312. Since the macro is only needed in that one case, would it make sense to just put the whole thing there?

BridgerVoss: It would also need to be called by md5-sha1 though, right?

padelsbach: Let's update the macro to include the null and void functions. Then we don't need the conditionals around the macro definition, similar to `IMPLEMENT_DIGEST`.

Like this:

```
#define IMPLEMENT_DIGEST_NULL(name)                                            \
static int wp_name##_digest_null(void) { return 0; }                           \
static void wp_name##_digest_void(void) {}                                     \
/** Dispatch table for digest algorithms. */                                   \
const OSSL_DISPATCH name##_functions[] = {                                     \
    { OSSL_FUNC_DIGEST_NEWCTX,          (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_INIT,            (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_UPDATE,          (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_FINAL,           (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_FREECTX,         (DFUNC)wp_name##_digest_void        }, \
    { OSSL_FUNC_DIGEST_DUPCTX,          (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_GET_PARAMS,      (DFUNC)wp_name##_digest_null        }, \
    { OSSL_FUNC_DIGEST_GETTABLE_PARAMS, (DFUNC)wp_name##_digest_null        }, \
    { 0,                                NULL                                }  \
};
```

PR #245, opened 2025-08-05
Update command line tests

padelsbach:
This update is needed for PR #234 

- Update simple.yml to use newer WP build process
- Remove build step from the `scripts/cmd_test` test suite to speed things up (by removing calls to `init_openssl` and `init_wolfssl`)
- Allow caller to provide their own openssl binary
- DRY out several blocks of repeated code

### DIFF
--- a/.github/workflows/simple.yml
+++ b/.github/workflows/simple.yml
@@ -13,25 +13,39 @@ concurrency:
 # END OF COMMON SECTION
 
 jobs:
+  build_wolfprovider:
+    uses: ./.github/workflows/build-wolfprovider.yml
+    with:
+      wolfssl_ref: ${{ matrix.wolfssl_ref }}
+      openssl_ref: ${{ matrix.openssl_ref }}
+    strategy:
+      matrix:
+        wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'master', 'openssl-3.5.0' ]
+        debug: ['WOLFPROV_DEBUG=1', '']
+
   simple_test:
     name: Simple Test
     runs-on: ubuntu-22.04
+    needs: build_wolfprovider
     timeout-minutes: 20
     strategy:
       matrix:
-        openssl_ref: [ 'master', 'openssl-3.5.0' ]
         wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'master', 'openssl-3.5.0' ]
         force_fail: ['WOLFPROV_FORCE_FAIL=1', '']
-        debug: ['WOLFPROV_DEBUG=1', '']
     steps:
       - name: Checkout wolfProvider
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
 
-      # Check if this version of wolfssl/wolfprovider has already been built,
-      # mark to cache these items on post if we do end up building
-      - name: Checking wolfSSL/wolfProvider in cache
-        uses: actions/cache@v4
-        id: wolfprov-cache
+      - name: Retrieving wolfProvider from cache
+        # Debug builds are not currently supported by build-wolfprovider.yml
+        # so those are manually built as a separate step.
+        if: ${{ matrix.debug == '' }}
+        uses: actions/cache/restore@v4
+        id: wolfprov-cache-restore
         with:
           path: |
             wolfssl-install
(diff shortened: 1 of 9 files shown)


.github/workflows/simple.yml
```diff
@@ -13,25 +13,35 @@ concurrency:
...
     runs-on: ubuntu-22.04
+    needs: build_wolfprovider
     timeout-minutes: 20
     strategy:
       matrix:
-        openssl_ref: [ 'master', 'openssl-3.5.0' ]
         wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
         force_fail: ['WOLFPROV_FORCE_FAIL=1', '']
-        debug: ['WOLFPROV_DEBUG=1', '']
```
aidangarske, line 25: Can we please keep the debug ref there has been some cases where debug has different behavior than normal.

.github/workflows/simple.yml
```diff
@@ -13,25 +13,35 @@ concurrency:
...
   simple_test:
     name: Simple Test
     runs-on: ubuntu-22.04
+    needs: build_wolfprovider
     timeout-minutes: 20
     strategy:
       matrix:
-        openssl_ref: [ 'master', 'openssl-3.5.0' ]
         wolfssl_ref: [ 'master', 'v5.8.0-stable' ]
+        openssl_ref: [ 'openssl-3.5.0' ]
```
aidangarske, line 34: any reason to remove master openssl ref?

scripts/cmd_test/do-cmd-tests.sh
```diff
@@ -19,42 +19,56 @@
...
-    echo "Please run build-wolfprovider.sh first"
-    exit 1
+    # Find the location of the wolfProvider modules
+    if [ -z "${WOLFPROV_PATH:-}" ]; then
+        export WOLFPROV_PATH=$(find /usr/lib /usr/local/lib -type d -name ossl-modules 2>/dev/null | head -n 1)
+    fi
+    # Set the path to the wolfProvider config file
+    if [ -z "${WOLFPROV_CONFIG:-}" ]; then
+        if [ "${WOLFSSL_ISFIPS:-0}" = "1" ]; then
+            export WOLFPROV_CONFIG="${REPO_ROOT}/provider-fips.conf"
```
aidangarske, line 49: Shouldn't this be set when we do the env-setup? / be able to be set there?

scripts/cmd_test/cmd-test-common.sh
```diff
@@ -0,0 +1,83 @@
...
+        exit 1
+    fi
+
+    # Fail flags
+    FAIL=0
+    FORCE_FAIL_PASSED=0
+
+    # Get the force fail parameter
+    if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
+        echo "Force fail mode enabled for RSA tests"
```
aidangarske, line 39: We should make this so it is dynamic with a var or just say "Force fail mode enabled" only once

scripts/cmd_test/do-cmd-tests.sh
```diff
@@ -19,42 +19,56 @@
...
-init_openssl
+# Ensure we can switch providers before proceeding
+use_default_provider
+use_wolf_provider
 
-echo "=== Running wolfProvider Command-Line Tests ==="
-echo "Using OpenSSL version: ${OPENSSL_TAG}"
-echo "Using wolfSSL version: ${WOLFSSL_TAG}"
+# Get the force fail parameter
+if [ "${WOLFPROV_FORCE_FAIL}" = "1" ]; then
```
aidangarske, line 66: Should we not call this again since we do this when we source the `cmd-test-common.sh`?

scripts/cmd_test/ecc-cmd-test.sh
```diff
@@ -220,6 +158,8 @@ test_sign_verify_pkeyutl() {
     fi
     
     # Test 3: Cross-provider verification (default sign, wolf verify)
+    # print FAIL
```
aidangarske, line 161: Why do we need to print the fail here?

aidangarske (commented):
Thanks for doing this looks a lot cleaner now

.github/workflows/simple.yml
padelsbach: Looks like ossl master has been removed from all other yamls. Should we keep testing it here?

scripts/cmd_test/cmd-test-common.sh
padelsbach: Good catch, will do

.github/workflows/simple.yml
aidangarske: We should honestly be testing this with more versions and not just 3.5.0 mainly for the make check. I would want master at least for the simple test.

.github/workflows/simple.yml
padelsbach: The cache tags don't identify the debug vs non-debug. Eg:

```
key: wolfprov-${{ matrix.wolfssl_ref }}-${{ matrix.openssl_ref }}-${{ github.sha }}
```

padelsbach: Latest push builds the debug version separately. Not optimal, but seems to work

.github/workflows/simple.yml
padelsbach: As discussed on Slack, tried this as a separate PR #247, but found issues with some combinations. Will postpone this work for now

scripts/cmd_test/ecc-cmd-test.sh
```diff
@@ -220,6 +158,7 @@ test_sign_verify_pkeyutl() {
     fi
     
     # Test 3: Cross-provider verification (default sign, wolf verify)
+    echo "Current FAIL status: $FAIL"
```
aidangarske, line 161: Do we still need this?

aidangarske: Looks good other than this

padelsbach: Good catch, will remove

aidangarske (approved):

PR #248, opened 2025-08-05
Fix WPFF runtime checks for key managment functions

aidangarske:
# Description

Added runtime validation and WPFF check in 
- DH - `wp_dh_dup()`, `wp_dh_decode()`, and `wp_dh_encode()`
- ECC  - `wp_ecc_decode_spki()`, `wp_ecc_decode_pki()`
- ECX  - `wp_ecx_dup()`, `wp_ecx_export()`, `wp_ecx_decode()`, `wp_ecx_encode()`
- MAC - `wp_mac_dup()`, `wp_mac_export()`, `wp_mac_gen_init()`
- RSA - `wp_rsa_dup()`, `wp_rsa_decode_spki()`, `wp_rsa_decode_pki()`, `wp_rsa_decode_enc_pki()`, `wp_rsa_encode()`, `wp_rsa_encode_text()`

### DIFF
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -490,14 +490,12 @@ static int wp_dh_copy_params(const wp_Dh *src, wp_Dh *dst)
  */
 static wp_Dh* wp_dh_dup(const wp_Dh *src, int selection)
 {
-    wp_Dh* dst;
+    wp_Dh* dst = NULL;
 
-    if (!wolfssl_prov_is_running()) {
-        return NULL;
+    if (wolfssl_prov_is_running()) {
+        /* Create a new dh object. */
+        dst = wp_dh_new(src->provCtx);
     }
-
-    /* Create a new DH key object to return. */
-    dst = wp_dh_new(src->provCtx);
     if (dst != NULL) {
         int ok = 1;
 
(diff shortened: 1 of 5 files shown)


src/wp_dh_kmgmt.c
```diff
@@ -492,6 +492,10 @@ static wp_Dh* wp_dh_dup(const wp_Dh *src, int selection)
 {
     wp_Dh* dst;
 
+    if (!wolfssl_prov_is_running()) {
```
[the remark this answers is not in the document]
padelsbach: Per discussion on Slack, early returns are not allowed in our code (but I see what you're going for!)

src/wp_rsa_kmgmt.c
```diff
@@ -3098,6 +3116,10 @@ static int wp_rsa_encode(wp_RsaEncDecCtx* ctx, OSSL_CORE_BIO* cBio,
     int ok = 1;
     int rc;
     BIO *out = wp_corebio_get_bio(ctx->provCtx, cBio);
```
padelsbach, line 3118: Probably doesn't make much of a difference, but we may want to avoid calling `wp_corebio_get_bio` until after checking wpff, for consistency.

Also, I think the check should go after all of the variables.

src/wp_rsa_kmgmt.c
```diff
@@ -4031,6 +4053,10 @@ static int wp_rsa_encode_text(wp_RsaEncDecCtx* ctx, OSSL_CORE_BIO* cBio,
 {
     int ok = 1;
     BIO *out = wp_corebio_get_bio(ctx->provCtx, cBio);
```
padelsbach, line 4055: Same comments here about `wp_corebio_get_bio` and the check following the variables.

src/wp_rsa_kmgmt.c
aidangarske: Ya that makes sense I was wondering that as well. Ill do that thanks

padelsbach (approved):
Looks good. Will give the others a chance to review before merging.

ColtonWilley (approved):

PR #249, opened 2025-08-06
n_fold: fix for carries (reverse add)

SparkiDev:
Also special handling when no adding required.
Improved test to check more constant lengths.

### DIFF
--- a/src/wp_krb5kdf.c
+++ b/src/wp_krb5kdf.c
@@ -292,84 +292,134 @@ static int wp_kdf_krb5kdf_expected_key_size(wp_Krb5kdfCtx* ctx)
  *   s[l] = (constant rot 13*(l/K))[l%k]
  *   block[l % N] += s[l] (with carry)
  * finally add carry if any
+ *
+ * @param [in] block         Block to fill with constant.
+ * @param [in] blocksize     Size of block in bytes.
+ * @param [in] constant      Constant to use for fill.
+ * @param [in] constant_len  Length of constant in bytes. Not zero and less
+ *                           than or equal to blocksize.
  */
 static void n_fold(unsigned char *block, unsigned int blocksize,
                    const unsigned char *constant, size_t constant_len)
 {
-    unsigned int cnt;
-    unsigned int i;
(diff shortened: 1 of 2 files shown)


src/wp_krb5kdf.c
```diff
@@ -292,84 +292,134 @@ static int wp_kdf_krb5kdf_expected_key_size(wp_Krb5kdfCtx* ctx)
...
+    unsigned int cnt;
     unsigned int carry;
-    unsigned int rot;
     unsigned int bi;
-    unsigned int const_len = (unsigned int)constant_len;
+    unsigned int rot;
+    unsigned int i;
+
+    /* Copy in unrotated constant first. */
+    XMEMCPY(block, constant, constant_len);
```
padelsbach, line 315: Should we validate `const_len` before doing this copy? (Not zero and less than or equal to blocksize)

src/wp_krb5kdf.c
```diff
@@ -292,84 +292,134 @@ static int wp_kdf_krb5kdf_expected_key_size(wp_Krb5kdfCtx* ctx)
...
-             /* Store new block value. */
-             block[bi] = (unsigned char)(carry & 0xff);
-             /* Get carry. */
-             carry >>= 8;
-
-             /* Next block index. */
-             bi = (bi + 1) % blocksize;
-         }
-         rot += 13;
+    for (i = cnt - const_len; i >= const_len; i -= const_len) {
```
padelsbach, line 375: Is it possible that `cnt - const_len` underflows or wraps?

src/wp_krb5kdf.c
SparkiDev: It is validated in callers.
It is documented in the function comments that constant_len is not zero and less than or equal to block size.

src/wp_krb5kdf.c
SparkiDev: No.
cnt is the lcm of constant_len and blocksize.
That is it is a multiple of both constant_len (const_len) and blocksize.

PR #251, opened 2025-08-06
Fix privSz for wp_dh_get_params

aidangarske:
# Description 

- `wp_dh_get_params` was returning the pub key size when it should return the private key size here. Seems like a non issue since the keys are usually the same size.


padelsbach:
Is there a test case we should add for this?

PR #258, opened 2025-08-12
Prevent python3-ntp and libhashkit workflows from running on every push

padelsbach:


padelsbach:
> So the changes are not to prevent them from running but to allow them to run on every push. Correct?

No, it's the other way around. It will now only run for pushes to master, main and release branches.

Also, a push to any branch with a PR will trigger it.

PR #259, opened 2025-08-13
Revamp Debug Selection and Output in wolfProvider

aidangarske:
# Description

- Added a  `WOLFPROV_ENTER` for every `WOLFPROV_LEAVE` in src... 
- Added new `WOLFPROV_MSG_DEBUG` and `WOLFPROV_MSG_TRACE` for higher levels of debug output
- Created bitmaps for `wolfProv_LogType` and `wolfProv_LogComponents` that allows us to select any pair of debug level and any pair of components.
- Created a certain `WP_LOG_QUERY` category that allows us to see the operation ID when we do a query
- Added debug for hkdf, kdf, and krb5 and any missing enter / leave debugging we didnt have in place

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ It will retrieve the dependencies and compile them as necessary. To use other th
 OPENSSL_TAG=openssl-3.5.0 WOLFSSL_TAG=v5.8.0-stable WOLFPROV_DEBUG=1 scripts/build-wolfprovider.sh
 ```
 
+Or you can set them with variables like so:
+
+```
+./scripts/build-wolfprovider.sh --debug --openssl-ver=openssl-3.5.0 --wolfssl-ver=v5.8.0-stable
+```
+
 To clean the build, use the following:
 ```
 ./scripts/build-wolfprovider.sh --clean
@@ -152,4 +158,9 @@ To run the command tests:
 
 To run the cipher suite testing:
 * `./scripts/test-wp-cs.sh`
+
+
+## Debugging
+
+To enable wolfProvider debug logging, build with `--debug`. If you want to filter logging a certain way, set `WOLFPROV_LOG_LEVEL_FILTER` and `WOLFPROV_LOG_COMPONENTS_FILTER` in `include/wolfprovider/wp_logging.h` as needed. See comments in that file for examples.
 
--- a/README.md
+++ b/README.md
@@ -162,5 +162,5 @@ To run the cipher suite testing:
 
 ## Debugging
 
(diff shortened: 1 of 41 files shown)


README.md
```diff
@@ -153,3 +159,8 @@ To run the command tests:
 To run the cipher suite testing:
 * `./scripts/test-wp-cs.sh`
 
+
+## Debugging
+
+To enable wolfProvider debug logging, build with `--debug`. If you want to filter logging a certain way, set `WOLFPROV_LOG_LEVEL_FILTER` and `WOLFPROV_LOG_COMPONENTS_FILTER` in `include/wolfprovider/wp_logging.h` as needed. See comments in that file for examples.
```
padelsbach, line 165: Will `--debug` turn on all logs? Might want to clarify

src/wp_logging.c
```diff
@@ -150,6 +150,10 @@ int wolfProv_SetLogComponents(int componentMask)
 static void wolfprovider_log(const int logLevel, const int component,
                            const char *const logMessage)
 {
+    /* Check compile-time configuration first */
+    if (!WOLFPROV_COMPILE_TIME_CHECK(component, logLevel))
```
padelsbach, line 154: Nit: I think we should have braces around these 3 `if` statements

src/wp_logging.c
```diff
@@ -225,6 +229,38 @@ void WOLFPROV_MSG_VERBOSE(int component, const char* fmt, ...)
...
+ * @param component [IN] Component type, from wolfProv_LogComponents enum.
+ * @param fmt   [IN] Log message format string.
+ * @param vargs [IN] Variable arguments, used with format string, fmt.
+ */
+WP_PRINTF_FUNC(2, 3)
+void WOLFPROV_MSG_DEBUG(int component, const char* fmt, ...)
+{
+    va_list vlist;
+    va_start(vlist, fmt);
+    wolfprovider_msg_internal(component, WP_LOG_DEBUG, fmt, vlist);
```
padelsbach, line 244: From my reading of the code, the logging code is not fully removed in non-debug. The logging functions still exist, do the va_args stuff, and even call sprintf. 

As part of this change, can you update it so that the logging functions (eg. `WOLFPROV_MSG_DEBUG`) don't exist in non-debug? 

Something like this in wp_logging.h
```
#define WOLFPROV_MSG_DEBUG(comp, fmt, ...) 
```

padelsbach (changes requested):

src/wp_logging.c
padelsbach: Update: this is already present in wp_logging.h, I just missed it. Carry on ;)

padelsbach (approved):
Looks good

ColtonWilley (approved):

PR #260, opened 2025-08-13
Initial option to replace openssl default provider

ColtonWilley:


default_stub/wp_default_stub.c
```diff
@@ -0,0 +1,40 @@
...
+ * along with wolfProvider. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <openssl/provider.h>
+
+/* Prototype of public function that initializes the wolfSSL provider. */
+OSSL_provider_init_fn wolfssl_provider_init;
+
+/* Prototype for the wolfprov_provider_init function */
+int wolfprov_provider_init(const OSSL_CORE_HANDLE* handle,
```
padelsbach, line 26: Nit: do we need the declaration if the actual function is just beneath?

default_stub/wp_default_stub.c
```diff
@@ -0,0 +1,40 @@
...
+                          const OSSL_DISPATCH* in,
+                          const OSSL_DISPATCH** out,
+                          void** provCtx);
+
+/*
+ * Provider implementation stub
+ */
+int wolfprov_provider_init(const OSSL_CORE_HANDLE* handle,
```
padelsbach, line 34: I think we should use the OSSL name `ossl_default_provider_init()`, etc. Then we can likely avoid the patch below. Should be easy to add the `_init()` functions here.

scripts/utils-openssl.sh
```diff
@@ -93,19 +93,74 @@ clone_openssl() {
...
+            else
+                printf "ERROR - stub libdefault not found in: ${STUB_LIB_DIR}\n"
+                do_cleanup
+                exit 1
+            fi
+        fi
+
+        # Execute configure
+        if [ "$WOLFPROV_REPLACE_DEFAULT" = "1" ]; then
+            $CONFIG_CMD LDFLAGS="${CONFIGURE_LDFLAGS}" LDLIBS="${CONFIGURE_LDLIBS}" >>$LOG_FILE 2>&1
```
padelsbach, line 155: Nit: could do this on line 155 instead of calling `$CONFIG_CMD` in two places. Would follow the pattern from above.

```
CONFIG_CMD+="LDFLAGS="${CONFIGURE_LDFLAGS}" LDLIBS="${CONFIGURE_LDLIBS}"
```

scripts/utils-wolfprovider.sh
```diff
@@ -46,6 +47,55 @@ WOLFPROV_DEBUG=${WOLFPROV_DEBUG:-0}
...
+        tail -n 40 $LOG_FILE
+        do_cleanup
+        exit 1
+    fi
+    printf "Done.\n"
+
+    cd ${SCRIPT_DIR}/..
+}
+
 clean_wolfprov() {
```
padelsbach, line 99: Can we add the default lib to the `clean_wolfprov()` function?

default_stub/wp_default_stub.c
ColtonWilley: I also thought it was overkill but clang was not happy otherwise.

.github/workflows/replace-default.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  replace_default_test:
+    name: Replace Default Test
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        # Test both standard and replace-default builds
+        replace_default: ['', '--replace-default']
```
padelsbach, line 23: Just curious: why not run with WPFF?

padelsbach (dismissed):
LGTM

patches/README.md
```diff
@@ -0,0 +1,4 @@
+This patch replaces the openssl default provider with wolfProvider. This means
```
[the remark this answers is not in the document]
ColtonWilley: You are right, and a link to the OSP patches as well. Very easy to think this is where the OSP patches are.

scripts/utils-openssl.sh
```diff
@@ -95,14 +95,45 @@ clone_openssl() {
     fi
 }
 
+check_openssl_replace_default_mismatch() {
+    # Only check if OpenSSL is already installed
+    if [ -d "${OPENSSL_INSTALL_DIR}" ]; then
+        local openssl_is_patched=0
+
+        # Check if the source was patched for --replace-default
+        if [ -f "${OPENSSL_SOURCE_DIR}/crypto/provider_predefined.c" ]; then
```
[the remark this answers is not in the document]
padelsbach: Maybe just add this additional check with a clear warning so that we know we're not testing the normal build.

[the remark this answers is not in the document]
ColtonWilley: I see, a nice big warning that you are not using stock openssl. I will add it.

padelsbach (approved):

PR #266, opened 2025-08-27
Fixes for 5.8.2 FIPS ready

ColtonWilley:
Fixes for 5.8.2 FIPS ready, including updated private key lock/unlock handling and updates for RSA min keysize change.


src/wp_ecc_kmgmt.c
```diff
@@ -733,7 +733,9 @@ static int wp_ecc_get_params_enc_pub_key(wp_Ecc* ecc, OSSL_PARAM params[],
                 outLen = 1 + 2 * ((ecc->bits + 7) / 8);
             }
             else {
+                PRIVATE_KEY_UNLOCK();
```
padelsbach, line 736: Consider putting these 3 lines in a function for DRYness, and to ensure the correct usage of `wc_ecc_export_x963_ex`.

Would have to take a function pointer as an arg though.

PR #268, opened 2025-08-28
Extending error printing for silent wolfssl errors (WIP)

JeremiahM37:
Added WOLFPROV_ERROR_MSG with the function name that is failing and the rc dump. Mostly done but still adding a few I missed or removing a few I shouldn't have added.


src/wp_aes_aead.c
```diff
@@ -869,6 +869,7 @@ static int wp_aesgcm_get_rand_iv(wp_AeadCtx* ctx, unsigned char* out,
 
         rc = wc_AesGcmInit(&ctx->aes, NULL, 0, ctx->iv, (word32)ctx->ivLen);
         if (rc != 0) {
+            WOLFPROV_MSG(WP_LOG_AES, "wc_AesGcmInit failed with rc=%d", rc);
```
aidangarske, line 872: Ideally this would be `WP_LOG_VERBOSE` or `WP_LOG_DEBUG` using the functions `WOLFPROV_MSG_VERBOSE` or `WOLFPROV_MSG_DEBUG`. Returning the rc for each function should be a higher level of debug then the normal `WOLFPROV_MSG` imo. I would use `WP_LOG_DEBUG` for this.

aidangarske (changes requested):

src/wp_aes_aead.c
JeremiahM37: Should they all be WP_LOG_DEBUG with WOLFPROV_MSG changed to WOLFPROV_MSG_DEBUG?

aidangarske: They are all the same so yes

aidangarske (commented):
You should find a place to mention that to see the rc from wolfssl you must add `WP_LOG_DEBUG` to the `WOLFPROV_LOG_LEVEL_FILTER`. Perhaps in `wp_logging.h` like in `wolfProv_LogType`

padelsbach (approved):
Nicely done

aidangarske (approved):

PR #272, opened 2025-09-03
Test debian bookworm for IGEL osp's

aidangarske:


.github/workflows/cjose.yml
```diff
@@ -68,18 +92,27 @@ jobs:
...
+          
+          # Set additional environment variables for OpenSSL
+          export OPENSSL_INSTALL=$GITHUB_WORKSPACE/openssl-install
+          export LDFLAGS="-L${OPENSSL_INSTALL}/lib64 -L${OPENSSL_INSTALL}/lib"
+          export CPPFLAGS="-I${OPENSSL_INSTALL}/include"
+          export PKG_CONFIG_PATH="${OPENSSL_INSTALL}/lib64/pkgconfig:${OPENSSL_INSTALL}/lib/pkgconfig:${PKG_CONFIG_PATH}"
+
           # Configure with OpenSSL
-          ./configure CFLAGS="-Wno-error=deprecated-declarations" --with-openssl=$GITHUB_WORKSPACE/openssl-install
+          ./configure CFLAGS="-Wno-error=deprecated-declarations" --with-openssl=$OPENSSL_INSTALL
```
ColtonWilley, line 107: For replace default model, I dont want to use explicit openssl install for configure. It should find system default. If that doesnt work with force fail, then this has not gone properly.

.github/workflows/curl.yml
```diff
@@ -59,16 +68,28 @@ jobs:
...
+          export DEBIAN_FRONTEND=noninteractive
           sudo apt-get update
-          sudo apt-get install nghttp2 libpsl5 libpsl-dev python3-impacket
+          sudo apt-get install -y build-essential autoconf automake libtool \
+              pkg-config nghttp2 libpsl5 libpsl-dev python3-impacket 
+
+      - name: Set up environment for curl build
+        run: |
+          # Set up the environment for wolfProvider
+          source $GITHUB_WORKSPACE/scripts/env-setup
```
ColtonWilley, line 79: With replace default, we should not use any sourcing or environment modification. We should ensure everything is installed into a system path and works with global defaults.

PR #277, opened 2025-09-10
Hostap workflow

JeremiahM37:
Mostly working hostap workflow with a patch for hostap hard loaded providers. Only running a small subset of the most essential tests. Force fail causes the tests to stall indefinitely, so the success case is that the tests timeout rather than actually fail.  Requires https://github.com/wolfSSL/osp/pull/300.

### DIFF
--- a/.github/workflows/hostap.yml
+++ b/.github/workflows/hostap.yml
@@ -1,0 +1,264 @@
+name: hostap and wpa supplicant Tests
+
+# START OF COMMON SECTION
+on:
+  push:
+    branches: [ 'master', 'main', 'release/**']
+  pull_request:
+    branches: [ '*' ]
+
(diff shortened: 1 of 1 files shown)


padelsbach:
@JeremiahM37, can you please rebase and see if the failing checks can be resolved by temporarily pointing to your fork?

padelsbach:
@JeremiahM37 I merged https://github.com/wolfSSL/osp/pull/300, can you please update this back to use OSP master?

JeremiahM37:
It seems the workflow is no longer working as expected after switching it to Debian, which is weird since I had no issues with it locally. 
image

The tests are falsely passing since the tests never run.

JeremiahM37:
Should be good now.

PR #286, opened 2025-09-22
Add git test for replace default

aidangarske:
# Description 

- Adds a `verify-debian.sh` flow that does a sanity check that we have correctly downloaded Debian wolfProvider
- Adds a script test `test-git-ssh-dr.sh` This test git with default replace algos incldue rsa, ecdsa, and ed25519 (can be ran locally with no changes)
- Added a yml github action that calls this called `git-ssh-dr.yml`
- did a refactor and implement the verify debian in all flows

### DIFF
--- a/.github/workflows/xmlsec.yml
+++ b/.github/workflows/xmlsec.yml
@@ -24,6 +24,7 @@ jobs:
         wolfssl_ref: [ 'v5.8.2-stable' ]
         openssl_ref: [ 'openssl-3.5.2' ]
         replace_default: [ true ]
+        fips: [ false ]
 
   test_xmlsec:
     runs-on: ubuntu-22.04
@@ -42,6 +43,7 @@ jobs:
         openssl_ref: [ 'openssl-3.5.2' ]
         force_fail: [ 'WOLFPROV_FORCE_FAIL=1', '' ]
         replace_default: [ true ]
+        fips: [ false ]
     env:
       WOLFSSL_PACKAGES_PATH: /tmp/wolfssl-packages
       OPENSSL_PACKAGES_PATH: /tmp/openssl-packages
@@ -80,6 +82,10 @@ jobs:
 
           apt install --reinstall -y \
             ${{ env.WOLFPROV_PACKAGES_PATH }}/libwolfprov_*.deb
+
+      - name: Verify wolfProvider is properly installed
+        run: |
+          $GITHUB_WORKSPACE/scripts/verify-install.sh ${{ matrix.replace_default && '--replace-default' || '' }} ${{ matrix.fips && '--fips' || '' }}
 
       - name: Install xmlsec dependencies
         run: |
@@ -122,8 +128,6 @@ jobs:
         shell: bash
         run: |
           set +o pipefail # ignore errors from make check
-          # Verify wolfProvider is properly installed
-          $GITHUB_WORKSPACE/scripts/verify-debian.sh
           export ${{ matrix.force_fail }}
           make check-keys | tee xmlsec-keys.log
           make check-enc | tee xmlsec-enc.log
--- a/.github/workflows/bind9.yml
+++ b/.github/workflows/bind9.yml
@@ -24,6 +24,7 @@ jobs:
         wolfssl_ref: [ 'v5.8.2-stable' ]
         openssl_ref: [ 'openssl-3.5.2' ]
         replace_default: [ true ]
+        fips: [ false ]
 
   test_bind:
     runs-on: ubuntu-22.04
@@ -42,6 +43,7 @@ jobs:
         openssl_ref: [ 'openssl-3.5.2' ]
         force_fail: ['WOLFPROV_FORCE_FAIL=1', '']
         replace_default: [ true ]
+        fips: [ false ]
     env:
       WOLFSSL_PACKAGES_PATH: /tmp/wolfssl-packages
       OPENSSL_PACKAGES_PATH: /tmp/openssl-packages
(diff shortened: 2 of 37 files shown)


scripts/verify-debian.sh
```diff
@@ -0,0 +1,147 @@
...
+    # Set environment variable for sourced usage
+    if [ $is_sourced -eq 1 ]; then
+        export WOLFPROVIDER_VERIFIED=$([ $overall_success -eq 0 ] && echo "1" || echo "0")
+        echo "Environment variable WOLFPROVIDER_VERIFIED set to: $WOLFPROVIDER_VERIFIED"
+    fi
+
+    return $overall_success
+}
+
+if [ $is_sourced -eq 0 ]; then
```
padelsbach, line 142: So if `is_sourced` is either 0 or 1 we run `verify_wolfprovider`. Is this logic correct?

scripts/verify-debian.sh
```diff
@@ -0,0 +1,147 @@
...
+        log_success "All wolfProvider verifications passed!"
+        echo "wolfProvider is correctly installed and configured on this Debian system."
+    else
+        handle_error "Some wolfProvider verifications failed!"
+        echo "wolfProvider is not properly installed on this Debian system."
+    fi
+
+    # Set environment variable for sourced usage
+    if [ $is_sourced -eq 1 ]; then
+        export WOLFPROVIDER_VERIFIED=$([ $overall_success -eq 0 ] && echo "1" || echo "0")
```
padelsbach, line 135: This value doesn't seem to be used. Is it needed?

.github/workflows/xmlsec.yml
```diff
@@ -123,6 +122,8 @@ jobs:
         shell: bash
         run: |
           set +o pipefail # ignore errors from make check
+          # Verify wolfProvider is properly installed
+          source $GITHUB_WORKSPACE/scripts/verify-debian.sh
```
padelsbach, line 126: Why `source` the script rather than run it? I think it's better to run it so that we don't pollute the callers env var namespace

scripts/test-git-ssh-dr.sh
```diff
@@ -0,0 +1,691 @@
...
+    print_status "SUCCESS" "Created bare repository at $TEST_BASE_DIR/test-repo.git"
+
+    # Create workspace and initial commit
+    mkdir test-workspace
+    cd test-workspace
+    git init
+    echo "# Test Repository" > README.md
+    git add README.md
+    git commit -m "Initial commit"
+    git remote add origin "$TEST_BASE_DIR/test-repo.git"
```
padelsbach, line 94: This creates a local repository right? I think testing our crypto against itself is good, but not ideal. Is there a way to test with another machine running standard/stock openssl+git?

scripts/verify-debian.sh
```diff
@@ -0,0 +1,147 @@
...
+log_success() {
+    echo "SUCCESS: $1"
+}
+log_info() {
+    echo "INFO: $1"
+}
+
+# Verify wolfProvider configuration file
+verify_provider_config() {
+    local provider_conf="/usr/lib/ssl/openssl.cnf.d/wolfprovider.conf"
```
padelsbach, line 57: I would argue that the existence of the conf file is not relevant, as long as `openssl list -providers` is correct

scripts/test-git-ssh-dr.sh
aidangarske: Are you talking about testing them with another repo or what do you mean by another machine how could we do this?

padelsbach: A simple access github.com via ssh and https should cover it:

```
git ls-remote git@github.com:wolfSSL/wolfProvider HEAD
git ls-remote https://github.com/git/git HEAD
```

The various git operations (clone, fetch, push, pull) all use the same crypto iiuc, so we might as well use a lightweight operation

scripts/verify-debian.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+}
+
+# Main verification function
+verify_wolfprovider() {
+    local overall_success=0
+
+    echo "=== wolfProvider Verification for Debian ==="
+    # Run all verification checks
+    verify_provider_loaded || overall_success=1
```
padelsbach, line 83: Can we add an argument to this script to select which of these checks to run? Eg. `--replace-default`

scripts/verify-debian.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo ""
+    verify_provider_version || overall_success=1
+    echo ""
+
+    echo "=== Verification Summary ==="
+    if [ $overall_success -eq 0 ]; then
+        log_success "All wolfProvider verifications passed!"
+        echo "wolfProvider is correctly installed and configured on this Debian system."
```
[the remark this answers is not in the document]
ColtonWilley: I hard agree on this one, we will be using the same flow to verify default replace behavior on non-debian systems

scripts/verify-debian.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo "OpenSSL version information:"
+    echo "$version_output"
+
+    if echo "$version_output" | grep -q "Library: OpenSSL 3.5.2+wolfProvider-nonfips"; then
+        log_success "wolfProvider replace default version is correct"
+        return 0
+    else
+        handle_error "wolfProvider replace default version is incorrect"
+        return 1
+    fi
```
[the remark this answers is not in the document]
ColtonWilley: Agree this is exactly what we should grep for, default + wolfSSL. We should also think about grepping for "active".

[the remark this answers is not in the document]
aidangarske: Agreed, Okay I will update that

scripts/verify-debian.sh
aidangarske: Yes I will do that as well.

padelsbach (approved):

ColtonWilley (approved):

ColtonWilley (approved):

PR #291, opened 2025-09-25
Fix is_openssl_patched

padelsbach:
- Fix build script function return values were not correct when using Debian source
- Remove GH action caching of Debian packages to avoid issues encountered with stale values (will revisit later)
- Use `--enable-wolfprovider` when building libwolfssl.deb to avoid test failures in OSPs
- More robust error handling in x11vnc yaml and test script
- Add newline in log in wolfprov init

### DIFF
--- a/src/wp_wolfprov.c
+++ b/src/wp_wolfprov.c
@@ -1242,8 +1242,9 @@ int wolfssl_provider_init(const OSSL_CORE_HANDLE* handle,
     if (ok) {
         if (wolfSSL_Debugging_ON() != 0) {
             WOLFPROV_MSG(WP_LOG_PROVIDER,
-              "WARNING: wolfProvider built with debug but underlying wolfSSL is not!"
-              "Building wolfSSl with debug is highly recommended, proceeding...");
+              "WARNING: wolfProvider built with debug but underlying wolfSSL is not!");
+            WOLFPROV_MSG(WP_LOG_PROVIDER,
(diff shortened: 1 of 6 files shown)


src/wp_wolfprov.c
```diff
@@ -1242,8 +1242,9 @@ int wolfssl_provider_init(const OSSL_CORE_HANDLE* handle,
     if (ok) {
         if (wolfSSL_Debugging_ON() != 0) {
             WOLFPROV_MSG(WP_LOG_PROVIDER,
-              "WARNING: wolfProvider built with debug but underlying wolfSSL is not!"
-              "Building wolfSSl with debug is highly recommended, proceeding...");
+              "WARNING: wolfProvider built with debug but underlying wolfSSL is not!");
```
padelsbach, line 1245: Unrelated, but I've been meaning to fix this

debian/install-wolfssl.sh
```diff
@@ -145,9 +146,16 @@ AC_CONFIG_FILES([debian/rules],[chmod +x debian/rules])' configure.ac
         --enable-keygen \
         --enable-shake128 \
         --enable-shake256 \
-        --enable-wolfprovider \
         --enable-rsapss \
-        --enable-scrypt"
+        --enable-scrypt \
```
padelsbach, line 150: These additions are WIP from testing without `--enable-wolfprovider`. Not sure if we want to add them now or wait for a full fix.

PR #294, opened 2025-09-26
Add FIPS version selection support

aidangarske:
# Description 

- Adds support for building different fips modules specifically `5.2.4` you can do this with normal `./scripts/build-wolfprovider.sh --enable-fips` which will default to 5.2.4 now and if you want to build with 5.2.1 or other you can set `--fips-version=5.2.1` for example
- [x] Needs PR [here](https://github.com/wolfSSL/fips-src/pull/87) first which fixes error with pilot check script


scripts/build-wolfprovider.sh
```diff
@@ -30,8 +30,8 @@ show_help() {
   echo "  WOLFSSL_TAG                wolfSSL tag to use (e.g., v5.8.0-stable)"
   echo "  WOLFSSL_ISFIPS             If set to 1, clones a wolfSSL FIPS bundle from GitHub"
   echo "  WOLFSSL_FIPS_BUNDLE        Directory containing the wolfSSL FIPS bundle to use instead of cloning from GitHub"
-  echo "  WOLFSSL_FIPS_VERSION       Version of wolfSSL FIPS bundle (v5, v6, ready), used as an argument for --enable-fips when configuring wolfSSL"
-  echo "  WOLFSSL_FIPS_CHECK_TAG     Tag for wolfSSL FIPS bundle (linuxv5.2.1, v6.0.0, etc), used as an argument for fips-check.sh when cloning a wolfSSL FIPS version"
+  echo "  WOLFSSL_FIPS_VERSION       Version of wolfSSL FIPS bundle (v5, v6, ready, 5.2.1, 5.2.4), used as an argument for --enable-fips when configuring wolfSSL"
```
padelsbach, line 33: What happens if `WOLFSSL_FIPS_VERSION` and `WOLFSSL_FIPS_CHECK_TAG` do not align?

Seems like it would simplify things a lot if we used _only_ the tag. What do you think?

aidangarske: I think you are right this would simplify it a lot

scripts/utils-wolfssl.sh
LinuxJedi, line 263: You probably want to run this is FIPS but not bundle? I think this might be causing your failures.

PR #345, opened 2025-12-24
Unify workflow yamls with pre-processing step

padelsbach:
- Use a new script to pre-process yamls and support `include:` directive to pull in shared code
- Generated files are checked in, and pre-processing ensured by new (non-generated) yaml
- Removes about 70 lines of common code and boilerplate setup

Notes on this approach: 
Pros:
- simple yaml-like syntax, still looks like workflow yaml
- shell script based, no additional installs (jinja2 was another option considered)

Cons:
- re-generation check is only done in CI
- `include` is only feature, no support for function-like macros or passing args


padelsbach:
This was mainly a proof of concept, with the main benefit of reducing duplicated code. Will revisit later, if needed.

PR #382, opened 2026-04-02
Initial implementation of SSHKDF for wolfprovider

ColtonWilley:
Initial implementation of SSHKDF for wolfprovider


test/test_sshkdf.c
```diff
@@ -0,0 +1,521 @@
...
+{
+    int err = 0;
+    int i;
+    unsigned char oKey[64];
+    unsigned char wKey[64];
+
+    /* Test input shared secret K. */
+    unsigned char inKey[] = {
+        0x00, 0x00, 0x00, 0x80,
+        0x55, 0xba, 0xe9, 0x31, 0xc0, 0x7f, 0xd8, 0x24,
```
padelsbach, line 85: Can we add a test case with the MSB set to test the sign extension path?

src/wp_sshkdf.c
```diff
@@ -0,0 +1,397 @@
...
+    if (ok) {
+        rawKey = ctx->key;
+        rawKeySz = (word32)ctx->keySz;
+
+        /* The caller passes K in SSH mpint encoding (4-byte big-endian length
+         * prefix, optional 0x00 padding byte, then value). wc_SSH_KDF() adds
+         * its own mpint encoding internally, so strip the caller's encoding
+         * to avoid double-encoding. */
+        if (rawKeySz >= 4) {
+            word32 mpintLen = ((word32)rawKey[0] << 24) |
```
padelsbach: wp_kbkdf.c uses this approach for endianness swap. Consider matching it here: 

```
/* We are not guaranteed to have these available from wolfssl, so implement
 * them here */
static void wp_c32toa(word32 wc_u32, byte* c) {
#ifdef WOLFSSL_USE_ALIGN
    c[0] = (byte)((wc_u32 >> 24) & 0xff);
    c[1] = (byte)((wc_u32 >> 16) & 0xff);
    c[2] = (byte)((wc_u32 >>  8) & 0xff);
    c[3] = (byte) (wc_u32 &        0xff);
#elif defined(LITTLE_ENDIAN_ORDER)
    *(word32*)c = ByteReverseWord32(wc_u32);
#else
    *(word32*)c = wc_u32;
#endif
```

test/test_sshkdf.c
ColtonWilley: added a test case

src/wp_sshkdf.c
ColtonWilley: We need the inverse but you are right. Added an equivalent helper and pulled them both to wp_internal.c

padelsbach (approved):

PR #387, opened 2026-04-06
Fix fenrir findings across wolfProvider

JeremiahM37:
# Description

Fixes F-174, F-175, F-176, F-186, F-187, F-188, F-260, F-261, F-262, F-263, F-272, F-273, F-274, F-400, F-399, F-505, F-506, F-507, F-508, F-509, F-831, F-832, F-833, F-834, F-1626, F-1627, F-1628, F-1635, F-1639, F-1640, F-1641, F-1642, F-1643

  - Chunk digest/HMAC/CMAC/AES/DES3 update calls to avoid size_t to word32 truncation on >4GB
  inputs                                                                                       
  - Fix SINGLE_THREADED → WP_SINGLE_THREADED mutex guards across all kmgmt files
  - Replace prefix-match XSTRNCMP with exact XSTRCMP in param/name lookups and fix sizeof      
  copy-paste errors                                                                            
  - Secure-clear key material in CMAC/GMAC free, EPKI decrypt, ECX key compare, and X25519     
  derive (constant-time)                                                                       
  - Guard tagLen/outLen assignment, KBKDF XMEMCPY NULL IV, GCM dinit NULL IV, and add RSA TLS
  implicit rejection                                                                           
  - Fix PEM footer length, base64 arithmetic, and hex dump off-by-one
  
 New Tests                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                           
  - test_digest_multi_update: SHA-256 single update vs. many 64-byte updates, verify match across OpenSSL and wolfProvider
  - test_hmac_multi_update: HMAC single update vs. many 128-byte updates, cross-provider comparison                                                                                                                                                                                                                 
  - test_cmac_multi_update: CMAC single update vs. many 16-byte updates, cross-provider comparison                                                                                                                                                                                                                  
  - test_aes_cbc_large_update: AES-256-CBC encrypt/decrypt roundtrip with chunked 1024-byte updates, cross-provider comparison                                                                                                                                                                                      
  - test_aes128_gcm_key_then_iv: GCM decrypt init with key only (NULL IV), then set IV via ctrl, encrypt/decrypt roundtrip      
 - test_rsa_param_prefix_match: Imports RSA key with prefix param "rsa-factor" —        
  incorrectly matches "rsa-factor1" on master                                                  
  - test_rsa_kem_prefix_match: Sets KEM operation to "RSASVE_extra" — incorrectly
  accepted on master

### DIFF
--- a/src/wp_cmac.c
+++ b/src/wp_cmac.c
@@ -91,7 +91,7 @@ static void wp_cmac_free(wp_CmacCtx* macCtx)
 {
     if (macCtx != NULL) {
         OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
-        OPENSSL_free(macCtx);
+        OPENSSL_clear_free(macCtx, sizeof(*macCtx));
     }
 }
 
@@ -222,14 +222,19 @@ static int wp_cmac_update(wp_CmacCtx* macCtx, const unsigned char* data,
     size_t dataLen)
 {
     int ok = 1;
-    int rc;
 
     WOLFPROV_ENTER(WP_LOG_COMP_MAC, "wp_cmac_update");
 
-    rc = wc_CmacUpdate(&macCtx->cmac, data, (word32)dataLen);
-    if (rc != 0) {
-        WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_CmacUpdate", rc);
-        ok = 0;
+    while (ok && (dataLen > 0)) {
+        word32 chunk = (dataLen > 0xFFFFFFFFU) ? 0xFFFFFFFFU : (word32)dataLen;
+        int rc = wc_CmacUpdate(&macCtx->cmac, data, chunk);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_CmacUpdate",
+                rc);
+            ok = 0;
+        }
+        data += chunk;
+        dataLen -= chunk;
     }
 
     WOLFPROV_LEAVE(WP_LOG_COMP_MAC, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
--- a/src/wp_cmac.c
+++ b/src/wp_cmac.c
@@ -90,6 +90,9 @@ static wp_CmacCtx* wp_cmac_new(WOLFPROV_CTX* provCtx)
 static void wp_cmac_free(wp_CmacCtx* macCtx)
 {
     if (macCtx != NULL) {
+    #ifndef HAVE_FIPS
+        wc_CmacFree(&macCtx->cmac);
+    #endif
         OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
         OPENSSL_clear_free(macCtx, sizeof(*macCtx));
     }
--- a/src/wp_rsa_asym.c
+++ b/src/wp_rsa_asym.c
@@ -466,27 +466,43 @@ static int wp_rsaa_decrypt(wp_RsaAsymCtx* ctx, unsigned char* out,
             if (ok) {
                 byte mask;
                 byte negMask;
+                byte rand[WOLFSSL_MAX_MASTER_KEY_LENGTH];
+                int i;
 
-                XMEMSET(out, 0, outSize);
-                PRIVATE_KEY_UNLOCK();
-                rc = wc_RsaPrivateDecrypt(in, (word32)inLen, out,
-                    (word32)outSize, wp_rsa_get_key(ctx->rsa));
-                PRIVATE_KEY_LOCK();
-
-                /* Constant time checking of master secret. */
-                mask  = wp_ct_byte_mask_eq(out[0], ctx->clientVersion >> 8);
-                mask &= wp_ct_byte_mask_eq(out[1], ctx->clientVersion);
-                if (ctx->negVersion > 0) {
-                    /* Check for negotiated version as well. */
-                    negMask  = wp_ct_byte_mask_eq(out[0], ctx->negVersion >> 8);
-                    negMask &= wp_ct_byte_mask_eq(out[1], ctx->negVersion);
-                    mask |= negMask;
-                }
-                rc &= (int)(char)mask;
-
-                if (rc <= 0) {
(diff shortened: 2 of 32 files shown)


aidangarske (commented):
Please don't just fix without adding test suites and verifying the actual failures here. We should write unit tests, validate these fail correctly, then fix so we know these are fixed and now tested.

JeremiahM37:
> Please don't just fix without adding test suites and verifying the actual failures here. We should write unit tests, validate these fail correctly, then fix so we know these are fixed and now tested.

The fixes in this PR are primarily defensive hardening, so direct testing is limited. Added tests that validate the fixed code paths.

dgarske (commented):
test body only

src/wp_cmac.c
```diff
@@ -91,7 +91,7 @@ static void wp_cmac_free(wp_CmacCtx* macCtx)
 {
```
dgarske, line 91: test inline

dgarske (commented):
test

test/unit.h
```diff
@@ -117,13 +117,17 @@ int test_shake_256(void *data);
 
 #endif /* WP_HAVE_DIGEST */
 
+int test_digest_multi_update(void *data);
```
dgarske, line 120: finding 1

dgarske (commented):
Skoll review

src/wp_rsa_asym.c
```diff
@@ -466,28 +466,44 @@ static int wp_rsaa_decrypt(wp_RsaAsymCtx* ctx, unsigned char* out,
...
+                            ctx->negVersion >> 8);
+                        negMask &= wp_ct_byte_mask_eq(out[1],
+                            ctx->negVersion);
+                        mask |= negMask;
+                    }
+                    /* Combine decrypt success with version check. */
+                    mask &= wp_ct_int_mask_gte(rc, 1);
+
+                    /* Constant-time select: real result or random fallback. */
+                    for (i = 0; i < WOLFSSL_MAX_MASTER_KEY_LENGTH; i++) {
```
dgarske, line 504: finding 2

src/wp_cmac.c
dgarske, line 91: finding 3

test/unit.h
dgarske, line 120: finding 1 body

src/wp_rsa_asym.c
dgarske, line 504: finding 2 body

src/wp_cmac.c
dgarske, line 91: finding 3 body

dgarske (changes requested):
overall body

src/wp_rsa_asym.c
dgarske, line 504: **🟡 [Medium] RSA TLS implicit rejection writes 48 bytes without defensive outSize check**
*💡 SUGGEST*  `question`

The implicit rejection code always writes `WOLFSSL_MAX_MASTER_KEY_LENGTH` (48) bytes to `out` in the constant-time select loop (line 501), regardless of `outSize`. While the size query path (line 414) returns 48, ensuring callers allocate enough, there is no defensive check that `outSize >= WOLFSSL_MAX_MASTER_KEY_LENGTH` before the loop. A misuse (e.g., caller passes a smaller buffer) would cause an out-of-bounds write. The old code had similar assumptions but the new 48-byte unconditional write makes it more explicit. This is the TLS pre-master secret path where 48 bytes is always expected, so it's low practical risk.

**Suggestion:**
```suggestion
Add a defensive check before the implicit rejection block:

if (ok && outSize < WOLFSSL_MAX_MASTER_KEY_LENGTH) {
    ok = 0;
}
```

dgarske (commented):
## 🐺 Skoll Code Review

**Overall recommendation:** `APPROVE`
**Findings:** 4 total — 3 posted, 1 skipped

### Posted findings
- [Medium] **test_digest_multi_update declared/registered outside WP_HAVE_DIGEST guard** — `test/unit.h:120, test/unit.c:205`
- [Medium] **RSA TLS implicit rejection writes 48 bytes without defensive outSize check** — `src/wp_rsa_asym.c:501`
- [Medium] **CMAC free does not call wc_CmacFree unlike GMAC's wc_AesFree addition** — `src/wp_cmac.c:90-95`

<details><summary>Skipped findings</summary>

- [Low] Inconsistent chunk max values could benefit from named constants

</details>

*Review generated by Skoll via openclaw*

src/wp_cmac.c
dgarske, line 91: **🟡 [Medium] CMAC free does not call wc_CmacFree unlike GMAC's wc_AesFree addition**
*💡 SUGGEST*  `convention`

The PR correctly added `wc_AesFree(&macCtx->gmac.aes)` in `wp_gmac_free` to properly release wolfSSL AES resources. However, the analogous `wp_cmac_free` (also modified in this PR to use `OPENSSL_clear_free`) does not call `wc_CmacFree` on the `Cmac` struct. wolfSSL's `wc_CmacFree` exists and is used in the codebase (e.g., `wp_kbkdf.c:521`). While the `OPENSSL_clear_free` zeroes memory, it doesn't run wolfSSL's cleanup for the internal AES state (which may hold hardware crypto handles on some platforms). Since this PR already touches `wp_cmac_free`, it would be consistent to add the wolfSSL cleanup call here too.

**Suggestion:**
```suggestion
static void wp_cmac_free(wp_CmacCtx* macCtx)
{
    if (macCtx != NULL) {
    #ifndef HAVE_FIPS
        wc_CmacFree(&macCtx->cmac);
    #endif
        OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
        OPENSSL_clear_free(macCtx, sizeof(*macCtx));
    }
}
```

src/wp_rsa_asym.c
JeremiahM37: added if (ok && (outSize < WOLFSSL_MAX_MASTER_KEY_LENGTH)) ok = 0; before the constant-time select loop.

src/wp_cmac.c
JeremiahM37: added wc_CmacFree(&macCtx->cmac) guarded by #ifndef HAVE_FIPS to match the GMAC pattern.

ColtonWilley (approved):

dgarske (approved):

PR #390, opened 2026-04-10
feat: route configurable devId through key init calls

MarkAtwood:
## Summary

wolfProvider currently hardcodes `INVALID_DEVID` in all key initialization
calls, silently bypassing wolfCrypt's device callback layer. This makes it
impossible to route OpenSSL operations through a wolfHSM server (or any
other wolfCrypt device callback) via the provider interface.

This PR closes that gap for RSA, ECC, and DH:

- Adds `int devId` to `WOLFPROV_CTX`, initialized to `INVALID_DEVID` (no behavior change for existing users)
- Exposes `wolfprovider_devid` as a settable `OSSL_PARAM` so callers can configure the device ID at runtime
- Routes `devId` through RSA (`wc_InitRsaKey_ex`), ECC (`wc_ecc_init_ex`), and DH (`wc_InitDhKey_ex`) key init calls
- Routes `devId` through the ECX gen-context RNG init (`wc_InitRng_ex`)

## Known gap

ECX key init functions (`wc_curve25519_init`, `wc_ed25519_init`, `wc_ed448_init`) use `WP_ECX_INIT` function pointers typed as `int(*)(void*)`, which carry no `devId` parameter. Routing devId through ECX key init requires changing the table shape and all four call sites — that is a larger, separate change. The RNG used during ECX key generation does use `devId` after this PR.

`wc_curve448_init_ex` does not exist in wolfCrypt as of wolfSSL 5.x, so curve448 cannot be fixed without a wolfCrypt change.

## Behavior for existing users

`devId` defaults to `INVALID_DEVID`, so all existing behavior is unchanged unless the caller explicitly sets `wolfprovider_devid` via `OSSL_PROVIDER_set_params`.

## Test plan

- [ ] Existing wolfProvider test suite passes (no regression)
- [ ] RSA sign/verify routes through wolfHSM device callback when `wolfprovider_devid` is set to `WH_DEV_ID`
- [ ] ECC sign/verify routes through wolfHSM device callback when `wolfprovider_devid` is set
- [ ] DH key agreement routes through wolfHSM device callback when `wolfprovider_devid` is set
- [ ] Setting `wolfprovider_devid` to `INVALID_DEVID` restores direct wolfCrypt behavior

### DIFF
--- a/include/wolfprovider/internal.h
+++ b/include/wolfprovider/internal.h
@@ -146,6 +146,7 @@ typedef struct WOLFPROV_CTX {
    wolfSSL_Mutex rng_mutex;
 #endif
    BIO_METHOD *coreBioMethod;
+   int devId;
 } WOLFPROV_CTX;
 
 #if defined(WP_HAVE_SEED_SRC) && defined(WP_HAVE_RANDOM)
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -396,7 +396,7 @@ static wp_Dh* wp_dh_new(WOLFPROV_CTX *provCtx)
         int ok = 1;
         int rc;
 
-        rc = wc_InitDhKey_ex(&dh->key, NULL, INVALID_DEVID);
+        rc = wc_InitDhKey_ex(&dh->key, NULL, provCtx->devId);
         if (rc != 0) {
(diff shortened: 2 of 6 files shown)


MarkAtwood:
Not need at this time

PR #391, opened 2026-04-13
Fix WPFF CI failures

padelsbach:
Bookworm updated to openssl 3.0.19 on 2026-04-07, right when CI started failing.

Workflows which run `apt install` after installing our patched openssl was now updating openssl, overwriting our version. The force-fail checks catch this and fail the CI run.

Complicating things, openssh now requires `libssl3 (>= 3.0.19)`.

Changelist:
- openssl's provider_predefined.c has minor whitespace changes, making the replace-default patch fail. replace the patch with a single drop-in file
- fix `install-openssl.sh` to check more `deb-src` locations to ensure we don't keep building 3.0.18, and enable 3.0.19 (or whatever is latest) to build via Jenkins
- apply `apt-mark hold` in CI anytime we install libwolfprov, libwolfssl or openssl to ensure it is not clobbered
- relatedly, check for held packages in addition to installed packages in `verify-install.sh`
- unrelated: force tnftp to be fetched from Debian instead of BSD for reliability


padelsbach:
revert changes here after merging: https://cloud.wolfssl-test.com/jenkins/job/wolfProvider/job/IGEL-openssl-debian12/configure

.github/workflows/bind9.yml
```diff
@@ -74,6 +74,11 @@ jobs:
           apt install --reinstall -y --allow-downgrades --allow-change-held-packages \
             ${{ env.WOLFPROV_PACKAGES_PATH }}/libwolfprov_*.deb
 
+          # Prevent later 'apt-get install' of test dependencies from
```
padelsbach, line 77: This ensures we don't clobber the installed libs in subsequent steps.

It's the same for all the yamls...

scripts/verify-install.sh
```diff
@@ -201,8 +201,10 @@ verify_wolfprovider() {
 
     detect_wolfprovider_mode
     dpkg_output=$(dpkg -l 2> /dev/null | grep wolf)
-    is_wolfssl_installed=$(echo "$dpkg_output" | grep -Eq '^ii\s+libwolfssl\s' && echo 1 || echo 0)
-    is_wolfssl_fips=$(echo "$dpkg_output" | grep -E '^ii\s+libwolfssl\s' | grep -qi "fips" && echo 1 || echo 0)
+    # Match any dpkg status whose second char is 'i' (installed state), so a
+    # package marked on hold ('hi') is still recognized as installed.
+    is_wolfssl_installed=$(echo "$dpkg_output" | grep -Eq '^[ih]i\s+libwolfssl\s' && echo 1 || echo 0)
```
padelsbach, line 206: Allows us to find the "held" package

padelsbach:
jenkins retest this please

ColtonWilley (approved):

PR #395, opened 2026-04-28
Misc kbkdf fixes

padelsbach:
1. F-3164: Increase buffer size to avoid a stack overflow possibility in `wp_kbkdf_init_hmac`.
2. F-2788: In `wp_kdf_kbkdf_set_ctx_params`, allow for explicitly set `OSSL_KDF_PARAM_MODE` field set to `"COUNTER"`. Prior to the fix, we support it if default/unspecified, but would not honor it when set specifically.

Added test cases for both.

### DIFF
--- a/src/wp_kbkdf.c
+++ b/src/wp_kbkdf.c
@@ -370,7 +370,10 @@ static void wp_c32toa(word32 wc_u32, byte* c) {
 }
 
 #ifdef WP_HAVE_HMAC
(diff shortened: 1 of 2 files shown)


padelsbach:
jenkins retest this please

padelsbach:
jenkins retest this please

ColtonWilley (approved):

PR #397, opened 2026-05-11
Add DH and ECDH test cases

padelsbach:
Fixes F-2228, F-2229, F-2232 (all missing tests reports)

### DIFF
--- a/test/test_dh.c
+++ b/test/test_dh.c
@@ -718,6 +718,149 @@ int test_dh_krb5_keygen(void *data)
     return err;
 }
 
+/* Generated offline with Python (g = 2):
+ *   p = int.from_bytes(dh_p, 'big')
+ *   pubA = pow(g, privA, p); pubB = pow(g, privB, p)
+ *   shared = pow(pubB, privA, p)
+ * privA was fixed and privB incremented until len(shared) < len(dh_p), i.e.
+ * the shared secret has a leading zero byte (a ~1/256 chance per try). Only
+ * privA, pubA (our key) and pubB (peer public) are needed at run time. */
+static const unsigned char dh_pad_privA[] = {
+    0xa3, 0x7f, 0x2c, 0x9e, 0x1b, 0x4d, 0x60, 0x58,
+    0xf3, 0xc2, 0xa1, 0x90, 0x7e, 0x5d, 0x4b, 0x3c,
+    0x2f, 0x19, 0x08, 0xe7, 0xd6, 0xc5, 0xb4, 0xa3,
+    0x92, 0x81, 0x70, 0x6f, 0x5e, 0x4d, 0x3c, 0x2b
+};
+
+static const unsigned char dh_pad_pubA[] = {
+    0x9c, 0x6c, 0xbc, 0xee, 0xdd, 0x3b, 0x2b, 0xc2,
+    0x47, 0x68, 0x71, 0xf1, 0xd2, 0x2d, 0x71, 0x95,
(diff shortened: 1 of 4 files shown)


padelsbach:
jenkins retest this please

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `COMMENT`

### Posted findings
- [Medium] [review] **DH padding test can fail nondeterministically** — `test/test_dh.c:786-901`
- [Medium] [review] **Undersized-buffer check masks successful truncated derives** — `test/test_dh.c:1178-1186`

test/test_dh.c
```diff
@@ -718,4 +718,472 @@ int test_dh_krb5_keygen(void *data)
...
+    EVP_PKEY_CTX *kgCtx = NULL;
+    EVP_PKEY *keyA = NULL;
+    EVP_PKEY *keyB = NULL;
+    unsigned char *secretPad = NULL;
+    unsigned char *secretNoPad = NULL;
+    size_t secretPadLen = 0;
+    size_t secretNoPadLen = 0;
+    const size_t maxLen = sizeof(dh_p);
+    int iter;
+    const int maxIter = 4096;
```
aidangarske: **🟠 [Medium] DH padding test can fail nondeterministically** · _test_

The new padding test depends on randomly generated DH key pairs eventually producing an unpadded shared secret shorter than the prime length. The PR then fails the test if that random event does not occur within `maxIter`. A correct provider can still fail this test, and the default suite now has variable runtime because it may perform thousands of 2048-bit DH key generations before hitting the case.

**Fix:** Make the padding regression deterministic so a correct implementation cannot fail due to random key generation. Use a deterministic DH key pair/vector whose shared secret has leading zero bytes with the existing `dh_p` and `dh_g`, or generate such a pair once and commit the fixed key material for this regression test.

test/test_dh.c
```diff
@@ -718,4 +718,472 @@ int test_dh_krb5_keygen(void *data)
...
+        if (test_dh_derive_with_x963(keyA, keyB, "SHA256", 32, NULL, 0,
+                tooSmallBuf, tooSmallLen) == 0) {
+            PRINT_ERR_MSG("DH X963 KDF derive accepted under-sized buffer");
+            err = 1;
+        }
+    }
+
+    OPENSSL_free(raw);
+    EVP_PKEY_CTX_free(kgCtx);
+    EVP_PKEY_free(keyA);
```
aidangarske, line 1274: **🟠 [Medium] Undersized-buffer check masks successful truncated derives** · _test_

The new negative test intends to verify that `EVP_PKEY_derive` fails when the caller buffer is smaller than the requested KDF output. It calls `test_dh_derive_with_x963`, but that helper converts any successful derive with `derivedLen != outLen` into an error. As a result, a provider bug that returns success while truncating to the 8-byte buffer would still make the helper return nonzero, and this test would pass even though the provider accepted the undersized buffer.

**Fix:** Adjust the negative test so it specifically fails if the provider returns success for an undersized output buffer. Build the derive context directly or add a helper mode that returns the raw `EVP_PKEY_derive` result, then assert that `EVP_PKEY_derive(ctx, tooSmallBuf, &tooSmallLen)` itself returns failure.

test/test_dh.c
padelsbach: done

test/test_dh.c
padelsbach: done

aidangarske (approved):

PR #400, opened 2026-05-23
Optimize CI for wolfProvider

aidangarske:
# Description 

Note 6/5: Currently CI is failing this will fix once merged

- trigger OSP projects to run nightly and send slack message if fail 
- dynamically get latest wolfssl version and openssl version 
- (All OSP where getting tested by 3.0.20 from debian:bookworm not 3.5.4)
- add ubsan and asan for WP specifically 
- Add smoke tests for draft 
- Only test on status "open" only smoke on draft 
- no apt-get use ghcr container 
- [x] backward comapt for 5.8.4
- [x] 5.9.1 support 
- Nightly runs all tested and passing with new patches
- Auto retry system for flaky nightlies in place and tested
- Added the ability to toggle pr runs of OSP projects on PR meaning we can add a label like `ci:all` or `ci:hostap` and it will run the selected suite in the PR you put the label on against that branch. Making it easy to test CI changes from failed nightly or add new tests or edit tests for example. 

related PR's need to go in first in this order then this one
1. https://github.com/wolfSSL/osp/pull/340
2. https://github.com/wolfSSL/testing/pull/962
3. https://github.com/wolfSSL/testing/pull/958

slack notifications system using claude and simple expected regex to retry flacky jobs
image


aidangarske:
Jenkins retest this please

aidangarske:
Jenkins retest this please

.github/workflows/hostap.yml
padelsbach: AI tells me this is a bug which will force the test to always pass. Unrelated, but since you're making changes here, maybe worth fixing?

.github/actions/wait-for-smoke/action.yml
```diff
@@ -0,0 +1,95 @@
+name: 'Wait for Smoke Test'
```
padelsbach: I'm not sure I see where/how this is used. Do the other tests wait for this to finish?

aidangarske: Yes this makes all the other tests wait to run only if the smoke test runs and passes like we do in wolfssl now.

ColtonWilley (approved):

PR #408, opened 2026-06-16
wp_drbg: gate wc_RNG_DRBG_Reseed on FIPS version

ColtonWilley:
wp_drbg_reseed references wc_RNG_DRBG_Reseed, but that is not always an exported symbol. It is WOLFSSL_API in non-FIPS >= 5.7.2 and in FIPS v6+, while FIPS v5.x commercial bundles are inconsistent -- the same v5.2.4 cert is WOLFSSL_LOCAL on a 5.8.4 wrapper but WOLFSSL_API on a 5.9.1 wrapper. Linking it fails (undefined reference to wc_RNG_DRBG_Reseed) on the bundles that keep it LOCAL, which the previous version-only gate did not account for.

- settings.h: gate WP_HAVE_DRBG_RESEED conservatively. Use the native in-place reseed only where the symbol is reliably exported (non-FIPS >= 5.7.2, FIPS v6+) and fall back to DRBG re-instantiation for all FIPS v5.x (the fallback links on every build). WP_NO_DRBG_RESEED forces the fallback for any other bundle that keeps the symbol WOLFSSL_LOCAL despite its version.
- wp_drbg.c: native in-place reseed when available, else re-instantiate the DRBG in place via wc_FreeRng()+wc_InitRngNonce() (with error-state reporting). Handle reseed with no caller entropy on both paths, and keep reseed seccomp-safe (native draws from the cached /dev/urandom fd via wp_urandom_read; fallback self-seeds via wc_InitRngNonce).
- utils-wolfssl.sh: map --fips-check=v5.2.4 to --enable-fips=v5.2.4 (was collapsing to v5, which mismatched the SP-math FIPS module).

Validated: build + full unit suite pass (136/0) against the 5.8.4 and 5.9.1 wolfSSL wrappers x v5.2.1 and v5.2.4 commercial FIPS bundles, plus non-FIPS.

Note on the FIPS v5.x fallback path: caller-supplied entropy/addIn is folded into wc_InitRngNonce as the nonce, not mixed as DRBG entropy_input, and predResist is not honored -- wolfCrypt re-seeds internally via its seed source. This is the available behavior where the module does not export wc_RNG_DRBG_Reseed, and differs from the native path where caller entropy is reseed input.

### DIFF
--- a/scripts/utils-wolfssl.sh
+++ b/scripts/utils-wolfssl.sh
@@ -179,7 +179,16 @@ install_wolfssl() {
             # Determine configure option from tag
             local fips_configure_arg=""
             case "$fips_tag" in
+                v5.2.4|linuxv5.2.4)
+                    # Distinct module (SP math, PATCH 4) — not the v5/cert4718 base
+                    fips_configure_arg="v5.2.4"
+                    ;;
+                v5.2.3|linuxv5.2.3)
+                    # Distinct module (PATCH 3) — not the v5/cert4718 base
+                    fips_configure_arg="v5.2.3"
+                    ;;
                 v5.2.*|v5.3.*|v5.4.*|v5.5.*|linuxv5.*)
+                    # v5.2.1/cert4718 and other v5.x map to the base v5 module
                     fips_configure_arg="v5"
                     ;;
                 v6.*|linuxv6.*)
@@ -320,7 +329,12 @@ install_wolfssl() {
                 do_cleanup
                 exit 1
             fi
-            cd ..
+            # Only the clone path descended into XXX-fips-test; bundles build in
+            # wolfssl-source directly. Must mirror the descent guard, else the
+            # bundle path pops one level too far.
+            if [ -z "$WOLFSSL_FIPS_BUNDLE" ]; then
+                cd ..
+            fi
             printf "Done.\n"
         fi
     fi
--- a/src/wp_drbg.c
+++ b/src/wp_drbg.c
@@ -27,6 +27,7 @@
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
 
+#include <wolfprovider/settings.h>
 #include <wolfprovider/alg_funcs.h>
 #include <wolfprovider/internal.h>
 
@@ -61,6 +62,10 @@ typedef struct wp_DrbgCtx {
     OSSL_FUNC_rand_get_seed_fn* parentGetSeed;
     /** Parent's clear_seed function. */
     OSSL_FUNC_rand_clear_seed_fn* parentClearSeed;
+#ifndef WP_HAVE_DRBG_RESEED
+    /** Set when a failed reseed re-instantiation left ctx->rng de-instantiated. */
+    int rngError;
+#endif
 } wp_DrbgCtx;
 
 
@@ -143,6 +148,8 @@ static void wp_drbg_free(wp_DrbgCtx* ctx)
     }
 }
 
+static int wp_drbg_uninstantiate(wp_DrbgCtx* ctx);
+
 /**
  * Instantiate a new DRBG.
  *
@@ -171,6 +178,11 @@ static int wp_drbg_instantiate(wp_DrbgCtx* ctx, unsigned int strength,
         ok = 0;
     }
 
+    /* Free any existing DRBG before re-allocating to avoid a leak. */
+    if (ok && ctx->rng != NULL) {
+        wp_drbg_uninstantiate(ctx);
+    }
+
     if (ok && ctx->parentGetSeed != NULL) {
         /* Get entropy from parent DRBG (no file I/O needed) */
         WOLFPROV_MSG_DEBUG(WP_LOG_COMP_RNG,
@@ -189,7 +201,9 @@ static int wp_drbg_instantiate(wp_DrbgCtx* ctx, unsigned int strength,
         }
 
         if (ok) {
-            /* Initialize wolfCrypt RNG with parent-provided seed */
+            /* Route DRBG instantiation through the FIPS-validated
+             * wc_InitRngNonce entry (not wc_rng_new), so it works on every
+             * pinned FIPS bundle vintage. */
             ctx->rng = OPENSSL_zalloc(sizeof(*ctx->rng));
             if (ctx->rng == NULL) {
                 ok = 0;
@@ -201,7 +215,7 @@ static int wp_drbg_instantiate(wp_DrbgCtx* ctx, unsigned int strength,
             if (rc != 0) {
                 WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_COMP_RNG,
                     "wc_InitRngNonce", rc);
-                OPENSSL_free(ctx->rng);
+                OPENSSL_clear_free(ctx->rng, sizeof(*ctx->rng));
                 ctx->rng = NULL;
                 ok = 0;
             }
@@ -242,6 +256,13 @@ static int wp_drbg_instantiate(wp_DrbgCtx* ctx, unsigned int strength,
     #endif
     }
 
+#ifndef WP_HAVE_DRBG_RESEED
+    if (ok) {
+        /* Clear any prior reseed error state. */
+        ctx->rngError = 0;
+    }
+#endif
+
     WOLFPROV_LEAVE(WP_LOG_COMP_RNG, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
     return ok;
 }
@@ -262,6 +283,9 @@ static int wp_drbg_uninstantiate(wp_DrbgCtx* ctx)
     OPENSSL_clear_free(ctx->rng, sizeof(*ctx->rng));
 #endif
     ctx->rng = NULL;
+#ifndef WP_HAVE_DRBG_RESEED
+    ctx->rngError = 0;
+#endif
     WOLFPROV_LEAVE(WP_LOG_COMP_RNG, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), 1);
     return 1;
 }
@@ -292,6 +316,16 @@ static int wp_drbg_generate(wp_DrbgCtx* ctx, unsigned char* out,
     if (strength > WP_DRBG_STRENGTH) {
         ok = 0;
     }
+    if (ok && ctx->rng == NULL) {
+        WOLFPROV_MSG_DEBUG(WP_LOG_COMP_RNG, "DRBG not instantiated");
+        ok = 0;
+    }
+#ifndef WP_HAVE_DRBG_RESEED
+    if (ok && ctx->rngError) {
+        WOLFPROV_MSG_DEBUG(WP_LOG_COMP_RNG, "DRBG in error state");
+        ok = 0;
+    }
+#endif
 #if 0
     if (ok && (addInLen > 0)) {
         rc = wc_RNG_DRBG_Reseed(ctx->rng, addIn, addInLen);
@@ -318,9 +352,11 @@ static int wp_drbg_generate(wp_DrbgCtx* ctx, unsigned char* out,
     return ok;
 }
 
-/* No usage of EVP_RAND_reseed seen in OpenSSL library. */
 /**
  * Reseed DRBG.
+ *
+ * Without WP_HAVE_DRBG_RESEED, re-instantiates instead of reseeding: @p entropy
+ * and @p addIn become the nonce, not DRBG entropy_input.
  *
  * @param [in, out] ctx         DRBG context object.
  * @param [in]      predResist  Prediction resistance required.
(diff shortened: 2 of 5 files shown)


src/wp_drbg.c
```diff
@@ -379,6 +437,52 @@ static int wp_drbg_reseed(wp_DrbgCtx* ctx, int predResist,
             ok = 0;
         }
     }
+#else
```
aidangarske, line 446: **🟠 [Medium] FIPS fallback reseed re-instantiates from OS entropy, bypassing the parent-seed/seccomp design and treating caller…** · _Security_

In the non-WP_HAVE_DRBG_RESEED fallback (FIPS modules without an exported wc_RNG_DRBG_Reseed, e.g. cert4718), reseed is implemented as wc_FreeRng(ctx->rng) + wc_InitRngNonce(ctx->rng, nonce, nonceLen), passing the caller-supplied entropy/addIn as the SP 800-90A NONCE rather than as entropy_input. This diverges from the native wc_RNG_DRBG_Reseed path in two ways: (1) wc_InitRngNonce always pulls fresh entropy from the OS seed source (wc_GenerateSeed -> getrandom()/dev/urandom) regardless of caller-supplied entropy, whereas the native path reseeds from caller entropy with no OS access; (2) it is a re-instantiation, not a true reseed, so the DRBG reseed counter/internal state semantics differ. The DRBG is not weakened (it is freshly seeded), and the divergence is documented in the function header. The notable risk worth confirming: wp_drbg_new documents that child DRBGs deliberately route entropy through the parent DRBG (parentGetSeed) so they avoid direct /dev/urandom access for seccomp-sandbox compatibility (see test/test_seccomp_sandbox.c). The in-place wc_InitRngNonce bypasses parentGetSeed entirely, so a reseed under a seccomp sandbox (reachable via RAND_seed/RAND_add, exercised by test_rand.c) could hit /dev/urandom and be blocked/killed on exactly the cert4718 FIPS builds this PR targets. Severity views: review rated Medium/SUGGEST (sandbox interaction worth confirming); review-security rated Info (documented, fails safe). Kept the stricter Medium.

**Fix:** Confirm (ideally with a sandbox test on a fallback/cert4718 build that triggers a reseed) that the fallback wc_InitRngNonce path does not break the seccomp-sandbox guarantee the parent-seed design provides. If it can, route the fallback re-instantiation seed through ctx->parentGetSeed when available, mirroring wp_drbg_instantiate, to preserve the no-direct-/dev/urandom invariant.

src/wp_drbg.c
```diff
@@ -61,6 +62,10 @@ typedef struct wp_DrbgCtx {
     OSSL_FUNC_rand_get_seed_fn* parentGetSeed;
     /** Parent's clear_seed function. */
     OSSL_FUNC_rand_clear_seed_fn* parentClearSeed;
+#ifndef WP_HAVE_DRBG_RESEED
```
aidangarske, line 65: **🟠 [Medium] New error-state and explicit-entropy reseed paths have no test coverage** · _Test Coverage_

The PR adds new state logic only exercised on FIPS-fallback builds: ctx->rngError being set on a failed re-instantiation, wp_drbg_generate refusing to run in the error state, and wp_drbg_get_ctx_params reporting EVP_RAND_STATE_ERROR/EVP_RAND_STATE_UNINITIALISED. The existing test/test_rand.c only calls RAND_seed/RAND_add/RAND_bytes on a healthy DRBG and never asserts on the reported OSSL_RAND_PARAM_STATE or drives a reseed failure, so none of the new error-handling branches are validated. make test passing on cert4718 only confirms the happy-path fallback works.

**Fix:** Add a test that (a) reads OSSL_RAND_PARAM_STATE via EVP_RAND_CTX_get_params and asserts EVP_RAND_STATE_READY after instantiate, and (b) forces a reseed failure to confirm the state transitions to EVP_RAND_STATE_ERROR and that subsequent generate calls fail, plus recovery via uninstantiate+instantiate.

src/wp_drbg.c
```diff
@@ -189,22 +201,29 @@ static int wp_drbg_instantiate(wp_DrbgCtx* ctx, unsigned int strength,
         }
 
         if (ok) {
-            /* Initialize wolfCrypt RNG with parent-provided seed */
-            ctx->rng = OPENSSL_zalloc(sizeof(*ctx->rng));
+            /* Use wc_rng_new (>= 5.0) so alloc and free use the same allocator,
+             * matching the root path and wc_rng_free() in teardown. */
+        #if LIBWOLFSSL_VERSION_HEX >= 0x05000000
+            ctx->rng = wc_rng_new(seed, (word32)seedLen, NULL);
             if (ctx->rng == NULL) {
```
aidangarske: **🔵 [Low] wp_drbg_get_seed does not check the new rngError state, operating on a de-instantiated RNG after a failed fallback…** · _State Management_

This PR introduces a new failure state in the non-WP_HAVE_DRBG_RESEED fallback of wp_drbg_reseed: if wc_InitRngNonce fails after wc_FreeRng(ctx->rng), ctx->rng is left non-NULL but de-instantiated (internal DRBG freed) and ctx->rngError = 1 is set. The PR correctly teaches wp_drbg_generate (line 328-333) and wp_drbg_get_ctx_params (line 629-634) to honor ctx->rngError, but wp_drbg_get_seed was NOT updated. It still gates only on if (ctx->rng == NULL) (line 734), so on a parent DRBG whose reseed re-instantiation failed, a child's GET_SEED request (OSSL_FUNC_RAND_GET_SEED) falls through to wc_RNG_GenerateBlock(ctx->rng, buffer, ...) (line 750) on a de-instantiated WC_RNG. Before this PR the reseed path used wc_RNG_DRBG_Reseed, which never de-instantiates, so ctx->rng != NULL always implied a usable RNG and the single NULL check was sufficient. wolfCrypt's wc_RNG_GenerateBlock guards a de-instantiated RNG (returns an error rather than dereferencing freed memory), so this fails safe with no crash, use-after-free, or weak-randomness leak — hence Low/Info — but the error-state handling is inconsistent with the other two consumers. Severity views: review-security rated Low (CWE-665, defense-in-depth); bugs rated Info (narrow trigger). Kept the stricter Low.

**Fix:** Mirror the guard used in wp_drbg_generate: immediately after the existing ctx->rng == NULL test in wp_drbg_get_seed, add an #ifndef WP_HAVE_DRBG_RESEED block: if (ctx->rngError) { WOLFPROV_MSG_DEBUG(WP_LOG_COMP_RNG, "DRBG in error state"); goto end; } so a failed-reseed parent reports failure to children consistently rather than relying on wolfCrypt's internal status guard.

src/wp_drbg.c
```diff
@@ -379,6 +437,52 @@ static int wp_drbg_reseed(wp_DrbgCtx* ctx, int predResist,
...
+            }
+        }
+        if (ok) {
+            XMEMCPY(nonce, entropy, entropyLen);
+            if (addIn != NULL && addInLen > 0) {
+                XMEMCPY(nonce + entropyLen, addIn, addInLen);
+            }
+        }
+        if (ok) {
+            wc_FreeRng(ctx->rng);
```
aidangarske, line 477: **🔵 [Low] wc_FreeRng() return code ignored in fallback reseed** · _Code Convention_

wc_FreeRng(ctx->rng) in the fallback branch discards its return value. The wolfSSL C coding standard calls for checking every return code. This matches the existing style in this file (wp_drbg_free/wp_drbg_uninstantiate also ignore wc_FreeRng's return in the <5.0 path), so it is consistent rather than a regression — noted only for completeness.

**Fix:** Optional: capture and debug-log the wc_FreeRng return code for consistency with the coding standard, or leave as-is to match the file's existing convention.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security + bugs`**Overall recommendation:** `COMMENT`
**Findings:** 5 total — 4 posted, 1 skipped

### Posted findings
- [Medium] [review+review-security] **FIPS fallback reseed re-instantiates from OS entropy, bypassing the parent-seed/seccomp design and treating caller entropy as nonce** — `src/wp_drbg.c:440-485`
- [Medium] [review] **New error-state and explicit-entropy reseed paths have no test coverage** — `src/wp_drbg.c:328-333,440-485,612-644`
- [Low] [review-security+bugs] **wp_drbg_get_seed does not check the new rngError state, operating on a de-instantiated RNG after a failed fallback reseed** — `src/wp_drbg.c:734-756`
- [Low] [review] **wc_FreeRng() return code ignored in fallback reseed** — `src/wp_drbg.c:468`

### Skipped findings
- [Low] ~~utils-wolfssl.sh adds untested v5.2.3 - --enable-fips=v5.2.3 mapping that changes prior behavior~~

src/wp_drbg.c
ColtonWilley: This is basically legacy path at this point, seems like test bloat rather than high ROI testing to me.

src/wp_drbg.c
ColtonWilley: added a test for this.

src/wp_drbg.c
ColtonWilley: Agreed, fix made

src/wp_drbg.c
ColtonWilley: leaving as is for consistency

scripts/utils-wolfssl.sh
```diff
@@ -179,7 +179,15 @@ install_wolfssl() {
             # Determine configure option from tag
             local fips_configure_arg=""
             case "$fips_tag" in
+                v5.2.4|linuxv5.2.4)
+                    # Distinct module (SP math, PATCH 4) — not the v5/cert4718 base
+                    fips_configure_arg="v5.2.4"
+                    ;;
+                v5.2.3|linuxv5.2.3)
```
aidangarske, line 186: **🟠 [Medium] Untested v5.2.3 FIPS tag mapping added without validation**

The PR adds a new case mapping `v5.2.3|linuxv5.2.3` to `--enable-fips=v5.2.3`. The PR description and validation only cover v5.2.1 and v5.2.4 bundles; v5.2.3 is neither mentioned nor tested, and unlike the v5.2.4 case it carries no comment explaining why it is a distinct module. If wolfSSL's `configure` does not accept `--enable-fips=v5.2.3` as a recognized value, builds invoked with that tag will now fail at configure time, whereas before this change they fell through to the working `v5` mapping. This silently broadens behavior for an unvalidated tag.

**Fix:** Confirm `--enable-fips=v5.2.3` is a valid configure value for the targeted wolfSSL versions. If it is not validated, drop the v5.2.3 case so it falls through to `v5` (the prior, working behavior), or add a comment documenting which cert/module it targets like the v5.2.4 case does.

src/wp_drbg.c
```diff
@@ -338,52 +381,130 @@ static int wp_drbg_reseed(wp_DrbgCtx* ctx, int predResist,
...
+
+    /* wolfCrypt RNG APIs take word32 lengths; reject oversized inputs. */
+    if (ok && entropy != NULL && entropyLen > 0xFFFFFFFFU) {
+        ok = 0;
+    }
+    if (ok && addIn != NULL && addInLen > 0xFFFFFFFFU) {
+        ok = 0;
+    }
+
+#ifdef WP_HAVE_DRBG_RESEED
```
aidangarske, line 393: **🟠 [Medium] Bare scope block introduced in wp_drbg_reseed native path**

The `#ifdef WP_HAVE_DRBG_RESEED` branch opens a standalone `{ ... }` block solely to scope the local declarations `unsigned char* seed` and `size_t seedLen`. This is a bare scope block that does not belong to a function, control statement, switch/case, or type definition. The wolfSSL/wolfProvider C style explicitly bans introducing bare scope blocks in new code to make C89 declarations work — declarations should move to the top of the enclosing function (or the logic split into a helper). Note the sibling `#else` fallback branch correctly declares its locals inside an `if (ok) {` control block, so the two halves of the same function are inconsistent.

**Fix:** Remove the bare scope block to comply with the no-bare-scope-block convention; the `empty-brace-scan` skill flags exactly this pattern.

scripts/utils-wolfssl.sh
```diff
@@ -179,7 +179,15 @@ install_wolfssl() {
             # Determine configure option from tag
             local fips_configure_arg=""
             case "$fips_tag" in
+                v5.2.4|linuxv5.2.4)
+                    # Distinct module (SP math, PATCH 4) — not the v5/cert4718 base
```
aidangarske, line 183: **🔵 [Low] Non-ASCII em-dash in shell comment**

The added comment uses a Unicode em-dash (`—`) in an otherwise ASCII source tree. For consistency with the rest of the C/shell sources and to avoid encoding surprises in tooling/diffs, prefer a plain ASCII hyphen.

**Fix:** Replace the em-dash with an ASCII hyphen.

src/wp_drbg.c
aidangarske, line 393: **🔵 [Low] Bare scope block introduced in wp_drbg_reseed native reseed path** · _Logic_

The new native (`WP_HAVE_DRBG_RESEED`) branch opens a standalone `{ ... }` scope solely to declare `unsigned char* seed` and `size_t seedLen` mid-function. This is a bare scope block, which the wolfSSL/wolfProvider C coding standard prohibits in new code (scope blocks are only allowed for function/control/switch/type bodies). The parallel fallback branch correctly uses a control-statement block (`if (ok) { ... }`), so only the native path violates the rule. BEFORE: these two locals were declared at the top of the function. AFTER: they were moved into a new bare `{ }` block under the `#ifdef`. This is a style/maintainability issue, not a security defect.

**Fix:** Declare `seed`/`seedLen` at the top of `wp_drbg_reseed` (guarded as needed) and remove the bare `{ }` scope, or move the native logic into a small helper function, to comply with the no-bare-scope coding standard. Run the empty-brace-scan to confirm no other new scope blocks were introduced.

aidangarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `COMMENT`
**Findings:** 6 total — 4 posted, 2 skipped

### Posted findings
- [Medium] [review] **Untested v5.2.3 FIPS tag mapping added without validation** — `scripts/utils-wolfssl.sh:186-188`
- [Medium] [review] **Bare scope block introduced in wp_drbg_reseed native path** — `src/wp_drbg.c:400-455`
- [Low] [review] **Non-ASCII em-dash in shell comment** — `scripts/utils-wolfssl.sh:183`
- [Low] [review-security] **Bare scope block introduced in wp_drbg_reseed native reseed path** — `src/wp_drbg.c:400-455`

### Skipped findings
- [Info] ~~FIPS v5.x fallback reseed deviates from SP 800-90A reseed semantics~~
- [Low] ~~Missing unit coverage for fallback reseed error-state path~~

aidangarske (approved):

PR #413, opened 2026-06-24
decoder: skip foreign-OID keys before FIPS CAST

ColtonWilley:
In FIPS, loading a key probes every keytype decoder and each instantiates a wolfCrypt key, firing that algorithm's lazy primitive-Z CAST (~150ms+) before any OID check. The decoder now checks the wrapped AlgorithmIdentifier OID and skips instantiation when the key is foreign and the CAST is still cold. Fixes the ~8s FIPS CMS/RSA-sign regression. Includes a FIPS regression test.


aidangarske (dismissed):

ColtonWilley:
retest this please.

aidangarske (approved):

PR #416, opened 2026-06-25
fix GCM tls iv length check and aead tag length validation

gasbytes:
Fixes the gcm tls fixed-iv length check to properly reject oversized inputs and adds tag length validation against each algorithm's allowed sizes for GCM and CCM, in particular this last one was wrapped in a new function called wp_aead_tag_len_valid.

### DIFF
--- a/src/wp_aes_aead.c
+++ b/src/wp_aes_aead.c
@@ -354,6 +354,47 @@ static int wp_aead_cache_in(wp_AeadCtx *ctx, const unsigned char *in,
 #endif
 
 /**
+ * Check whether a tag length is one of the algorithm's allowed sizes.
+ *
+ * GCM tags must be 4, 8, 12, 13, 14, 15 or 16 bytes (NIST SP 800-38D).
+ * CCM tags must be 4, 6, 8, 10, 12, 14 or 16 bytes (NIST SP 800-38C /
+ * RFC 3610).
+ *
+ * Note: for GCM this is intentionally stricter than OpenSSL's default
+ * provider, which accepts any non-zero tag length up to 16 on
+ * EVP_CTRL_AEAD_SET_TAG and EVP_CTRL_AEAD_GET_TAG. wolfProvider enforces the
+ * NIST SP 800-38D set on both the set path (decrypt/verify) and the get path
+ * (encrypt/tag retrieval), so tag lengths such as 5, 6, 7, 9, 10 or 11 that
+ * stock OpenSSL would accept are rejected here.
+ *
+ * @param [in] mode  Cipher mode: EVP_CIPH_GCM_MODE or EVP_CIPH_CCM_MODE.
+ * @param [in] sz    Tag length in bytes to check.
+ * @return  1 if sz is an allowed length for mode.
+ * @return  0 otherwise.
+ */
+static int wp_aead_tag_len_valid(int mode, size_t sz)
+{
+    int ok;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_AES, "wp_aead_tag_len_valid");
+
+    if (mode == EVP_CIPH_GCM_MODE) {
+        ok = (sz == 4) || (sz == 8) || (sz == 12) || (sz == 13) ||
+             (sz == 14) || (sz == 15) || (sz == 16);
+    }
+    else if (mode == EVP_CIPH_CCM_MODE) {
+        ok = (sz == 4) || (sz == 6) || (sz == 8) || (sz == 10) ||
+             (sz == 12) || (sz == 14) || (sz == 16);
+    }
+    else {
+        ok = 0;
+    }
+
+    WOLFPROV_LEAVE(WP_LOG_COMP_AES, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
+    return ok;
+}
+
+/**
  * Get the AEAD context parameters.
  *
  * @param [in]      ctx     AEAD context object.
@@ -432,7 +473,8 @@ static int wp_aead_get_ctx_params(wp_AeadCtx* ctx, OSSL_PARAM params[])
         if (p != NULL) {
             size_t sz = p->data_size;
             if ((!ctx->enc) || (ctx->tagLen == UNINITIALISED_SIZET) ||
-                (sz == 0) || (sz > ctx->tagLen)) {
+                (sz == 0) || (sz > ctx->tagLen) ||
+                (!wp_aead_tag_len_valid(ctx->mode, sz))) {
                 ok = 0;
             }
             if (ok && (!OSSL_PARAM_set_octet_string(p, ctx->buf, sz))) {
@@ -486,6 +528,9 @@ static int wp_aead_set_param_tag(wp_AeadCtx* ctx,
         sz = p->data_size;
     }
     if (ok && ((sz == 0) || ((p->data != NULL) && ctx->enc))) {
+        ok = 0;
+    }
+    if (ok && !wp_aead_tag_len_valid(ctx->mode, sz)) {
         ok = 0;
     }
     if (ok) {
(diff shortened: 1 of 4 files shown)


src/wp_aes_aead.c
```diff
@@ -458,6 +458,37 @@ static int wp_aead_get_ctx_params(wp_AeadCtx* ctx, OSSL_PARAM params[])
...
+ * @param [in] mode  Cipher mode: EVP_CIPH_GCM_MODE or EVP_CIPH_CCM_MODE.
+ * @param [in] sz    Tag length in bytes to check.
+ * @return  1 if sz is an allowed length for mode.
+ * @return  0 otherwise.
+ */
+static int wp_aead_tag_len_valid(int mode, size_t sz)
+{
+    int ok;
+
+    if (mode == EVP_CIPH_GCM_MODE) {
```
ColtonWilley: **🟠 [Medium] GCM tag-length validation is stricter than OpenSSL on decrypt (interop divergence)**

For GCM, `wp_aead_tag_len_valid` only accepts tag sizes {4,8,12,13,14,15,16} and `wp_aead_set_param_tag` now rejects anything else, including on the decrypt (verify) path. OpenSSL's own GCM provider does NOT enforce this discrete set on `EVP_CTRL_AEAD_SET_TAG`; it accepts any non-zero size up to 16 (e.g. 5,6,7,9,10,11). For CCM this exactly matches OpenSSL (4..16 even), so only GCM diverges. An application that previously set a non-standard GCM tag length (legal in stock OpenSSL) will now get a failure from wolfProvider. This is almost certainly the intended, more NIST-correct behavior, but it is a behavioral/interop divergence worth confirming and noting in a changelog.

**Fix:** Confirm intent; if intended (likely), document the divergence from OpenSSL's GCM tag-length acceptance so downstream consumers are aware.

test/test_aestag.c
```diff
@@ -1416,6 +1416,159 @@ int test_aes_gcm_bad_tag(void *data)
...
+    return err;
+}
+
+int test_aes_gcm_tag_len_undersized(void *data)
+{
+    int err;
+
+    (void)data;
+
+    PRINT_MSG("AES-128-GCM undersized tag length rejection");
```
ColtonWilley: **🟠 [Medium] New validation paths only partially exercised by tests**

The tests cover the headline cases (1-byte tag rejected for GCM/CCM; oversized fixed IV rejected for GCM) but leave gaps for the new logic: (1) only AES-128 variants are exercised, whereas sibling tests in this file (e.g. test_aes_gcm_bad_tag) also cover 256-bit; (2) no positive assertion that a *valid* non-default tag length (e.g. 12/13/14 for GCM, 8 for CCM) is still accepted, so a regression that over-rejects allowed sizes would go undetected; (3) for GCM, the 'in-between' sizes that the new function specifically excludes (e.g. 5/6/7) are not tested; (4) the fixed-IV test does not cover the `len == ctx->ivLen` boundary (which must be rejected because it leaves <8 bytes for the explicit field).

**Fix:** Extend the helpers to also assert accepted lengths and a couple of additional rejected lengths/key sizes to lock in the full validation table.

src/wp_aes_aead.c
```diff
@@ -458,6 +458,37 @@ static int wp_aead_get_ctx_params(wp_AeadCtx* ctx, OSSL_PARAM params[])
...
+ * GCM tags must be 4, 8, 12, 13, 14, 15 or 16 bytes (NIST SP 800-38D).
+ * CCM tags must be 4, 6, 8, 10, 12, 14 or 16 bytes (NIST SP 800-38C /
+ * RFC 3610).
+ *
+ * @param [in] mode  Cipher mode: EVP_CIPH_GCM_MODE or EVP_CIPH_CCM_MODE.
+ * @param [in] sz    Tag length in bytes to check.
+ * @return  1 if sz is an allowed length for mode.
+ * @return  0 otherwise.
+ */
+static int wp_aead_tag_len_valid(int mode, size_t sz)
```
ColtonWilley: **🔵 [Low] New helper omits WOLFPROV_ENTER/LEAVE tracing used by sibling functions**

Nearly every other static helper in this file (e.g. `wp_aead_set_param_tag`, `wp_aead_set_param_iv_len`, `wp_aead_cache_aad`) opens with `WOLFPROV_ENTER(WP_LOG_COMP_AES, "...")` and returns via `WOLFPROV_LEAVE`. The newly added `wp_aead_tag_len_valid` has neither. It is a trivial pure predicate so the omission is harmless, but it is inconsistent with the established convention.

**Fix:** Optional: add tracing to match the file's convention, or accept the omission for a pure helper.

test/test_aestag.c
```diff
@@ -1416,6 +1416,159 @@ int test_aes_gcm_bad_tag(void *data)
...
+    unsigned char tag[16];
+    int outLen = 0, fLen = 0;
+
+    memset(key, 0xAA, keyLen);
+    memset(iv, 0xBB, sizeof(iv));
+
+    cipher = EVP_CIPHER_fetch(libCtx, cipherName, "");
+    err = cipher == NULL;
+
+    /* Encrypt normally to get a valid ciphertext/full-size tag pair. */
```
ColtonWilley: **🔵 [Low] Undersized-tag tests perform an unnecessary full encrypt preamble**

Both undersized-tag helpers run a complete encrypt + GET_TAG sequence solely to populate the `tag` buffer, but the behavior under test (rejecting a 1-byte tag length at `EVP_CTRL_AEAD_SET_TAG`) is rejected before any tag bytes are consumed, so the encryption is not needed. This couples the rejection test to unrelated encrypt-path success: if CCM/GCM encryption ever regresses, these tests would fail for a reason unrelated to tag-length validation, obscuring the signal.

**Fix:** Simplify the helpers to test only the rejection path, decoupling them from encrypt-path success.

ColtonWilley (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `COMMENT`
**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] **GCM tag-length validation is stricter than OpenSSL on decrypt (interop divergence)** — `src/wp_aes_aead.c:473-490, 522`
- [Medium] **New validation paths only partially exercised by tests** — `test/test_aestag.c:1474-1485, 1559-1570, 1847-1858`
- [Low] **New helper omits WOLFPROV_ENTER/LEAVE tracing used by sibling functions** — `src/wp_aes_aead.c:473-490`
- [Low] **Undersized-tag tests perform an unnecessary full encrypt preamble** — `test/test_aestag.c:1511-1535, 1779-1820`

src/wp_aes_aead.c
gasbytes: Thanks, addressed.

src/wp_aes_aead.c
gasbytes: Thanks, addressed.
I added a small note saying that we know diverge a little bit from OpenSSL by being stricter on the which tags we accept for AEAD with AES-GCM.

test/test_aestag.c
gasbytes: Thanks, addressed.

test/test_aestag.c
gasbytes: Thanks, addressed.

src/wp_aes_aead.c
```diff
@@ -488,6 +529,9 @@ static int wp_aead_set_param_tag(wp_AeadCtx* ctx,
     if (ok && ((sz == 0) || ((p->data != NULL) && ctx->enc))) {
```
aidangarske, line 530: **🟠 [High] Tag length validation is bypassed when retrieving AEAD tags**
*🚫 BLOCK*

The PR adds `wp_aead_tag_len_valid()` and wires it only into the set-tag path. For encryption, applications choose the returned tag length through `EVP_CTRL_AEAD_GET_TAG`, which goes through `wp_aead_get_ctx_params()` and still accepts any nonzero `sz <= ctx->tagLen`. After a normal GCM encrypt, a caller can still request a 1-byte tag and the provider will return it, even though this PR is intended to reject tag sizes outside the algorithm's allowed set. The new tests only exercise `EVP_CTRL_AEAD_SET_TAG` on decrypt, so this public API path remains uncovered.

**Recommendation:** Apply `wp_aead_tag_len_valid(ctx->mode, sz)` in `wp_aead_get_ctx_params()` when handling `OSSL_CIPHER_PARAM_AEAD_TAG`, then extend the new tag-length tests to cover the get-tag/encryption path. Add regression tests that `EVP_CTRL_AEAD_GET_TAG` rejects invalid lengths such as 1, 5, and 11 bytes while accepting allowed lengths.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Tag length validation is bypassed when retrieving AEAD tags** — `src/wp_aes_aead.c:522`

*Review generated by Skoll.*

src/wp_aes_aead.c
gasbytes: Thanks! Missed that test gap. Should be addressed now.

aidangarske:
Jenkins Retest this please

aidangarske (approved):

ColtonWilley (approved):

PR #417, opened 2026-06-29
fix wp_ecx_get_security_bits to return the correct security bits for x448 and x25519

gasbytes:
fix wp_ecx_get_security_bits threasholds so x25519 returns 128 and x448 returns 224

### DIFF
--- a/src/wp_ecx_kmgmt.c
+++ b/src/wp_ecx_kmgmt.c
@@ -511,11 +511,15 @@ static int wp_ecx_get_security_bits(wp_Ecx* ecx)
 {
     int bits = 0;
 
-    if (ecx->data->bits >= 456) {
-        bits = 224;
-    }
-    else if (ecx->data->bits >= 256) {
-        bits = 128;
+    switch (ecx->data->keyType) {
+        case WP_KEY_TYPE_X448:
+        case WP_KEY_TYPE_ED448:
+            bits = 224;
+            break;
+        case WP_KEY_TYPE_X25519:
+        case WP_KEY_TYPE_ED25519:
+            bits = 128;
+            break;
     }
 
     return bits;
--- a/test/test_ecx.c
+++ b/test/test_ecx.c
@@ -26,11 +26,11 @@
 #include <wolfssl/wolfcrypt/ed25519.h>
 #include <wolfssl/wolfcrypt/ed448.h>
 
-#if defined(WP_HAVE_ED25519) || defined(WP_HAVE_ECD448)
-
 #ifndef ARRAY_SIZE
     #define ARRAY_SIZE(a)   ((sizeof(a)/sizeof(a[0])))
 #endif
+
+#if defined(WP_HAVE_ED25519) || defined(WP_HAVE_ED448)
 
 #ifndef MAX
     #define MAX(a,b)       ((a) > (b) ? (a) : (b))
@@ -656,7 +656,6 @@ static int test_ecx_null_sign_init_ex(OSSL_LIB_CTX *libCtx)
 
 #ifndef WP_HAVE_ED25519
     (void)libCtx;
-    (void)provider_name;
     PRINT_MSG("Skipping test - WP_HAVE_ED25519 not defined");
     return 0;
 #endif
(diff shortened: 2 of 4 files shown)


test/test_ecx.c
```diff
@@ -897,3 +897,79 @@ int test_ecx_dup(void *data)
...
+    } types[] = {
+#ifdef WP_HAVE_X25519
+        { "X25519" },
+#endif
+#ifdef WP_HAVE_X448
+        { "X448" },
+#endif
+    };
+
+    for (unsigned i = 0; i < ARRAY_SIZE(types) && err == 0; i++) {
```
ColtonWilley, line 926: **🔴 [High] New test uses ARRAY_SIZE outside its defining feature guard — compile break in curve-only builds**

The new test `test_ecx_x_security_bits` is wrapped in `#if defined(WP_HAVE_X25519) || defined(WP_HAVE_X448)` (line 900) and calls `ARRAY_SIZE(types)` at line 926. However, `ARRAY_SIZE` is `#define`d only inside the `#if defined(WP_HAVE_ED25519) || defined(WP_HAVE_ECD448)` block at lines 29-33, which closes at line 898 — before the new test begins. Since a `#define` is only processed when its enclosing `#if` is true, `ARRAY_SIZE` is available only when `WP_HAVE_ED25519` is defined (`WP_HAVE_ECD448` is a typo that is never defined; the macro is `WP_HAVE_ED448`). In `include/wolfprovider/settings.h` (lines 174-184), `WP_HAVE_X25519`/`WP_HAVE_X448` (from `HAVE_CURVE25519`/`HAVE_CURVE448`) are gated independently of `WP_HAVE_ED25519`/`WP_HAVE_ED448`. So a build with curve25519/curve448 enabled but ed25519/ed448 disabled compiles this test with `ARRAY_SIZE` undefined, producing a compile error (`'ARRAY_SIZE' undeclared`). `ARRAY_SIZE` is not provided by any included wolfSSL header. Default builds that enable Ed25519 compile fine, which is why this is easy to miss.

**Fix:** Hoist the `ARRAY_SIZE` definition above the `#if defined(WP_HAVE_ED25519) ...` guard (or add a local `#ifndef ARRAY_SIZE` inside the new block) so the X25519/X448-only build configuration compiles. Verify with a curve-only build (e.g. wolfSSL `--enable-curve25519 --disable-ed25519 --disable-ed448`).

ColtonWilley (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 3 total — 3 posted, 0 skipped
**3 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **New test uses ARRAY_SIZE outside its defining feature guard — compile break in curve-only builds** — `test/test_ecx.c:926`
- [Medium] **Security-bits test only checks parity with OpenSSL, not the expected absolute values** — `test/test_ecx.c:951-958`
- [Low] **Magic-number thresholds in wp_ecx_get_security_bits are undocumented** — `src/wp_ecx_kmgmt.c:514-518`

test/test_ecx.c
```diff
@@ -897,3 +897,79 @@ int test_ecx_dup(void *data)
...
+        }
+        if (err == 0) {
+            pkey_ossl = NULL;
+            err = EVP_PKEY_generate(ctx_ossl, &pkey_ossl) != 1;
+        }
+        if (err == 0) {
+            pkey_wolf = NULL;
+            err = EVP_PKEY_generate(ctx_wolf, &pkey_wolf) != 1;
+        }
+        if (err == 0) {
```
ColtonWilley, line 951: **🟠 [Medium] Security-bits test only checks parity with OpenSSL, not the expected absolute values**

The test validates correctness only by comparing wolfProvider's `EVP_PKEY_get_security_bits` against OpenSSL's (`sec_ossl != sec_wolf`). The PR's stated intent is that X25519 returns 128 and X448 returns 224. The differential check transitively verifies this only because OpenSSL happens to return those values; it would not catch a case where both sides are wrong, and it pins the test to OpenSSL's behavior rather than the documented expectation. This matches the differential-testing convention used elsewhere in this file, so it is acceptable, but asserting the absolute expected value per curve would make the test directly exercise the threshold logic that was changed in `wp_ecx_get_security_bits`.

**Fix:** Optionally assert the absolute expected security bits (128 for X25519, 224 for X448) in addition to the OpenSSL parity check, so the test directly pins the threshold logic this PR changed.

src/wp_ecx_kmgmt.c
```diff
@@ -511,10 +511,10 @@ static int wp_ecx_get_security_bits(wp_Ecx* ecx)
 {
     int bits = 0;
 
-    if (ecx->data->bits >= 456) {
+    if (ecx->data->bits >= 448) {
```
ColtonWilley: **🔵 [Low] Magic-number thresholds in wp_ecx_get_security_bits are undocumented**

The thresholds `>= 448` and `>= 255` are derived from the `data->bits` values assigned to each curve (x25519=255, x448=448, ed25519=256, ed448=456), but that relationship is non-obvious to a reader. The function infers a security level from a key-size threshold even though `data->keyType` (WP_KEY_TYPE_X25519, etc.) is available and would express intent more directly. The pre-existing code already used magic numbers, so this matches existing style, but a short comment mapping the thresholds to the curve families would help future maintainers and prevent regressions like the off-by-a-few-bits values this PR is fixing.

**Fix:** Add a brief comment explaining how the 255/448 thresholds map to the X/Ed curve families, or switch on `data->keyType` for clarity. Non-blocking.

test/test_ecx.c
gasbytes: Thanks, fixed!

src/wp_ecx_kmgmt.c
gasbytes: Thanks, fixed! Changed to a switch instead of adding the comment.

test/test_ecx.c
```diff
@@ -26,12 +26,12 @@
 #include <wolfssl/wolfcrypt/ed25519.h>
 #include <wolfssl/wolfcrypt/ed448.h>
 
-#if defined(WP_HAVE_ED25519) || defined(WP_HAVE_ECD448)
-
 #ifndef ARRAY_SIZE
     #define ARRAY_SIZE(a)   ((sizeof(a)/sizeof(a[0])))
 #endif
 
+#if defined(WP_HAVE_ED25519) || defined(WP_HAVE_ED448)
```
aidangarske, line 33: **🟠 [High] ED448-only test build now reaches undeclared provider_name**
*🚫 BLOCK*  `bug`

The PR changes the outer ECX test guard from the non-existent `WP_HAVE_ECD448` to `WP_HAVE_ED448`, which correctly enables this block for Ed448-only builds. That newly exposed configuration compiles `test_ecx_null_sign_init_ex()` without `WP_HAVE_ED25519`, and the skip branch references `provider_name`, which is not declared in this translation unit. `grep` only found `provider_name` as a separate static in `src/wp_wolfprov.c`, so a build with `WP_HAVE_ED448` and no `WP_HAVE_ED25519` will fail to compile the unit test.

**Recommendation:** Remove the undeclared `provider_name` reference, or otherwise make the Ed448-only test path compile cleanly now that the PR enables it.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **ED448-only test build now reaches undeclared provider_name** — `test/test_ecx.c:33`

*Review generated by Skoll.*

ColtonWilley (approved):

aidangarske (approved):

PR #424, opened 2026-07-01
ci: ORAS-backed OpenSSL/wolfSSL dep cache for from-source builders

aidangarske:
## Description

Adds an ORAS-backed dependency cache for the from-source CI builders. OpenSSL/wolfSSL
are built once per resolved commit SHA and pushed to `ghcr.io/wolfssl/wolfprovider/build-deps`,
then pulled by every from-source job — the same registry pattern the repo already uses for
the `debs:` packages. Addresses the PR #412 review: durable, cross-run, cross-workflow reuse
instead of relying only on the evictable GitHub Actions cache.

The build scripts already skip configure+make when `openssl-install`/`wolfssl-install` exist
(`utils-openssl.sh`, `utils-wolfssl.sh`), so a pulled artifact is a drop-in — no build-script
changes needed.

### How the layers work (per job)

1. Resolve openssl/wolfssl ref → commit SHA (`resolve-ref.sh`), compute tags keyed by
   `variant + SHA + build-script-hash`.
2. `oras pull` the install dirs. Hit → build skips.
3. On a miss, fall back to `actions/cache` (PR #412's key).
4. Build only what is still absent.
5. On an ORAS miss where the dir now exists, `oras push` it so the next job/run hits warm
   (self-healing; canonical repo + write token only).

Master refs are keyed by resolved HEAD SHA, so master is rebuilt only when upstream master
actually advances — not every run.

### New

- `scripts/oras-build-deps.sh` — pull/push helper.
- `.github/actions/oras-build-deps` — pull + GH-cache fallback.
- `.github/actions/oras-build-deps-push` — self-healing push.

### Wired (14 from-source workflows)

cmdline, simple, smoke-test, seed-src, fips-ready, multi-compiler, nightly-multi-compiler,
sanitizers (isolated asan/tsan variants), openssl-version, wolfssl-versions-pqc,
wolfssl-pqc-kat, libtss2, sssd, static-analysis.

`wolfssl-pqc-kat.yml` is also reverted from the temporary fork branch back to upstream refs.

### Deployment note

First canonical run creates `ghcr.io/wolfssl/wolfprovider/build-deps` (private). An org admin
must flip it to **Public** so fork PRs can pull anonymously (same as the `debs:` package). Until
then forks fall back to build — no broken state.

### DIFF
--- a/.github/actions/oras-build-deps/action.yml
+++ b/.github/actions/oras-build-deps/action.yml
@@ -1,0 +1,238 @@
+name: ORAS build deps
+description: Pull prebuilt OpenSSL/wolfSSL installs from ghcr.io (ORAS) keyed by resolved commit SHA, falling back to actions/cache, so build-wolfprovider.sh skips configure+make on a hit.
+
+inputs:
+  variant:
+    description: Token capturing every build-config dimension that changes the produced binaries (compiler, debug, replace-default, seed-src, fips). Distinct configs must use distinct variants. Installs bake in an absolute prefix/rpath, so a variant must also be unique per runner image - do not reuse a variant across the container/non-container boundary (e.g. multi-compiler uses -bookworm, nightly uses -${OS}).
+    required: true
+  openssl_ref:
+    description: OpenSSL tag/branch/sha, resolved to a SHA for the artifact key.
+    required: true
+  wolfssl_ref:
+    description: wolfSSL tag/branch/sha, resolved to a SHA for the artifact key. Ignored when wolfssl_fixed_key is set.
+    required: false
+    default: ''
+  wolfssl_fixed_key:
+    description: Verbatim wolfSSL key, used instead of resolving wolfssl_ref (FIPS bundle case - fixed stable version with no git ref).
+    required: false
+    default: ''
+  cache_openssl_source:
+    description: Also fetch openssl-source. Required for --replace-default / --fips-baseline builds whose mismatch checks read the patched source.
+    required: false
+    default: 'false'
+  extra_key:
+    description: Extra key segment folded into the artifact tag and fallback cache key. Use for build flags defined in the workflow file (e.g. sanitizer CFLAGS) so a flag change invalidates the artifact.
+    required: false
+    default: ''
+  github_token:
+    description: Token passed to resolve-ref.sh and oras login to avoid GitHub API rate limits.
+    required: true
+
+outputs:
+  registry:
+    description: ghcr.io OCI repo the artifacts live in.
+    value: ${{ steps.tags.outputs.registry }}
+  openssl_install_tag:
+    description: Tag for the openssl-install artifact.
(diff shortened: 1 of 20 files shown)


.github/actions/oras-build-deps/action.yml
```diff
@@ -0,0 +1,238 @@
...
+      shell: bash
+      run: |
+        if "${GITHUB_WORKSPACE}/scripts/oras-build-deps.sh" pull \
+             "${{ steps.tags.outputs.registry }}:${{ steps.tags.outputs.wolfssl_install_tag }}" wolfssl-install; then
+          echo "hit=true" >> "$GITHUB_OUTPUT"
+        else
+          echo "hit=false" >> "$GITHUB_OUTPUT"
+        fi
+
+    # Fallback: when ORAS misses, restore PR #412's actions/cache so a warm
```
padelsbach: Nit: can we remove the reference to other PRs?

.github/actions/oras-build-deps/action.yml
```diff
@@ -0,0 +1,238 @@
...
+    # being present (same replace-default coupling as the ORAS pulls).
+    - name: Cache OpenSSL source (fallback)
+      id: src-fallback
+      if: inputs.cache_openssl_source == 'true' && steps.pull-openssl-source.outputs.hit != 'true'
+      uses: actions/cache@v4
+      with:
+        path: openssl-source
+        key: openssl-source-${{ inputs.variant }}-${{ steps.openssl-ref.outputs.sha }}-${{ hashFiles('scripts/utils-*.sh', 'scripts/build-wolfprovider.sh', 'scripts/patch-*.sh', 'patches/**') }}
+
+    - name: Cache OpenSSL install (fallback)
```
padelsbach, line 226: IIUC, both the built `openssl-source` and `openssl-install` are needed because the "lives in openssl-source/test and only exists after a full source build". 

In that case, we might want to simplify to a single openssl source+install entry.

aidangarske: They're coupled for the cache_openssl_source=true rows (replace-default / FIPS / PQC-KAT), but I kept them as two tags on purpose the openssl-install tag is deliberately shared across jobs with different source needs. In pqc-kat, for example, the pre-PQC [build-only] row (no source) and the [non-replace] PQC rows (need source) resolve to the same openssl-install tag, since stock OpenSSL is identical regardless of the wolfSSL/pqc build. Folding source into that artifact would either force the build-only row (and every non-cache_openssl_source builder) to pull a source tree it discards, or leave one tag with two different contents which just reintroduces the coupling. Keeping openssl-source as its own tag, pulled only by the rows that set cache_openssl_source, is what keeps the shared install lean and safe to use the openssl-install

padelsbach: ok

padelsbach (approved):

PR #425, opened 2026-07-01
fips: define DH_MIN_SIZE for the FIPS PILOT build

aidangarske:
## Description

The FIPS PILOT build (`--enable-fips=v5`, `fips-check-PILOT.sh`) fails to compile:

```
wolfcrypt/src/dh.c: 'DH_MIN_SIZE' undeclared (first use in this function)
    if (mp_count_bits(&key->p) < DH_MIN_SIZE)   // GeneratePrivateDh / GeneratePublicDh / wc_DhAgree_Sync
```

### Root cause

The PILOT builds `BASE_TAG=v5.2.1-stable` and overlays FIPS-module crypto files.
The module `dh.c` now carries a `< DH_MIN_SIZE` bounds check, but the
`v5.2.1-stable` base `settings.h` predates the `DH_MIN_SIZE` definition (added to
wolfSSL master only recently), so the symbol is undeclared in that build.

The PILOT runs `./configure --enable-fips=$FIPS_OPTION` with no `CFLAGS`, so the
build inherits flags from the environment.

### Fix

Supply the FIPS `DH_MIN_SIZE` value (2048, matching master's `settings.h`
`HAVE_FIPS` branch) via `CPPFLAGS`, scoped to the PILOT invocation. `CPPFLAGS`
(not `CFLAGS`) is used so autoconf's default `-g -O2` is preserved — important so
the FIPS in-core integrity hash is computed against the normal optimized build.

Scoped to the PILOT path only; the `linuxv5.*`/`linuxv6.*` bundles are
self-consistent and unaffected.

### DIFF
--- a/scripts/utils-wolfssl.sh
+++ b/scripts/utils-wolfssl.sh
@@ -259,6 +259,8 @@ install_wolfssl() {
                 # So for the 'git' commands, we'll just use whatever the system comes with.
                 if [ "$fips_check_script" = "fips-check-PILOT.sh" ]; then
                     # PILOT script has different usage: [flavor] [keep]
+                    # v5.2.1 FIPS base predates DH_MIN_SIZE; export it for the PILOT and the XXX-fips-test rebuilds below.
+                    export CPPFLAGS="${CPPFLAGS:-} -DDH_MIN_SIZE=2048"
                     LD_LIBRARY_PATH="" ./$fips_check_script "$fips_tag" keep >$LOG_FILE 2>&1
                     RET_CODE=$?
                 else
--- a/scripts/utils-wolfssl.sh
+++ b/scripts/utils-wolfssl.sh
@@ -259,9 +259,9 @@ install_wolfssl() {
                 # So for the 'git' commands, we'll just use whatever the system comes with.
                 if [ "$fips_check_script" = "fips-check-PILOT.sh" ]; then
                     # PILOT script has different usage: [flavor] [keep]
-                    # v5.2.1 FIPS base predates DH_MIN_SIZE; export it for the PILOT and the XXX-fips-test rebuilds below.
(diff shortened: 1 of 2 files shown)


scripts/utils-wolfssl.sh
```diff
@@ -259,7 +259,9 @@ install_wolfssl() {
                 # So for the 'git' commands, we'll just use whatever the system comes with.
                 if [ "$fips_check_script" = "fips-check-PILOT.sh" ]; then
                     # PILOT script has different usage: [flavor] [keep]
-                    LD_LIBRARY_PATH="" ./$fips_check_script "$fips_tag" keep >$LOG_FILE 2>&1
+                    # v5.2.1 FIPS base predates DH_MIN_SIZE; add it to the FIPS wolfSSL build only - WOLFSSL_CONFIG_CFLAGS feeds the XXX-fips-test rebuilds, CPPFLAGS the PILOT's own configure.
```
padelsbach: Nit: this comment is quite long

test/test_rsa.c
```diff
@@ -1192,9 +1192,12 @@ int test_rsa_pkey_keygen(void *data)
      * we're using wolfCrypt FIPS. Can't do 2048 because that's the default. */
     const int newKeySize = 3072;
     const int badKeyGenSizes[] = {512, 1024};
+    /* FIPS 186 requires the RSA public exponent >= 65537; e=3 is rejected. */
+    const long newKeyExp = 65537;
```
padelsbach, line 1196: Could be useful to check other values here to confirm the fix. Eg:

```
const int badKeyGenExps[] =  {3, 65535};
```

And maybe check another good value too, eg `65539`.

Edit: maybe this belongs in wolfcrypt (as well?). AI tells me that there is no "too small" test there either.

padelsbach (approved):

PR #427, opened 2026-07-03
Enforce DH minimum prime size on the parameters-only generation path

yosuke-wolfssl:
## Summary

DH parameter generation could produce keys weaker than the provider's own declared minimum on the **parameters-only** path.

`wp_dh_params_validate` — which enforces `WP_DH_MIN_BITS` — was only invoked when `OSSL_KEYMGMT_SELECT_KEYPAIR` was set. A caller requesting only `OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS` (parameter-only generation) could therefore have the provider generate DH parameters below the minimum with no check, and `wp_dh_gen_set_params` read `OSSL_PKEY_PARAM_FFC_PBITS` into `ctx->bits` with no range check.

This is most impactful in **FIPS** builds: `WP_DH_MIN_BITS` is 2048, but wolfCrypt's own `wc_DhGenerateParams` floor is 1024, so sub-2048 parameters could be generated unchecked. In non-FIPS builds wolfCrypt's floor (1024) already equals `WP_DH_MIN_BITS`, so there the change is defense-in-depth.

## Changes

- **`wp_dh_gen`**: run `wp_dh_params_validate` after parameters are generated or copied, in **every** selection path, before key pair generation (previously only under the keypair branch).
- **`wp_dh_gen_set_params`**: reject an `FFC_PBITS` request below `WP_DH_MIN_BITS` at set-params time (fail-fast for callers that check the return).
- **`test_dh_pgen_min_bits`** (new): asserts a below-minimum prime length is rejected and a request at exactly the minimum still succeeds. Gated on `HAVE_FIPS` to mirror `WP_DH_MIN_BITS` (1024 non-FIPS / 2048 FIPS).

## Testing

- Full DH unit suite passes (tests 77–86), including the new `test_dh_pgen_min_bits`.
- Negative control: disabling the size guard makes the new test fail at the set-params assertion, confirming it is non-vacuous.

## Notes

- `WP_DH_MIN_BITS` is left at 1024 for non-FIPS to avoid breaking interop with legitimate 1024-bit DH consumers. Raising it to 2048 (per NIST SP 800-131A / Logjam) is a separate policy decision and not included here.
- The relocated `wp_dh_params_validate` is the actual enforcer; OpenSSL buffers `prime_len` and the set-params guard is a secondary fail-fast.

### DIFF
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -51,13 +51,6 @@ static const int wp_dh_decode_nids[] = {
 /** Maximum size of the group name string. */
 #define WP_MAX_DH_GROUP_NAME_SZ     10
 
-/* Min accepted bitlen for keygen */
-#ifdef HAVE_FIPS
-#define WP_DH_MIN_BITS 2048
-#else
-#define WP_DH_MIN_BITS 1024
-#endif
-
 /**
  * DH key.
  */
@@ -1574,17 +1567,18 @@ static wp_DhGenCtx* wp_dh_gen_init(WOLFPROV_CTX* provCtx,
             WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_InitRng", rc);
             ok = 0;
         }
-        if (ok) {
-            if (!wp_dh_gen_set_params(ctx, params)) {
-                wc_FreeRng(&ctx->rng);
-                ok = 0;
-            }
-        }
+        /* Defaults first so init-time parameters override them. */
         if (ok) {
             ctx->provCtx   = provCtx;
             ctx->selection = selection;
             ctx->bits      = 2048;
             ctx->generator = 2;
+        }
+        if (ok) {
+            if (!wp_dh_gen_set_params(ctx, params)) {
+                wc_FreeRng(&ctx->rng);
+                ok = 0;
+            }
         }
 
         if (!ok) {
@@ -1644,13 +1638,25 @@ static int wp_dh_gen_set_template(wp_DhGenCtx* ctx, const wp_Dh* dh)
 static int wp_dh_gen_set_params(wp_DhGenCtx* ctx, const OSSL_PARAM params[])
 {
     int ok = 1;
+    int bits;
     const OSSL_PARAM* p;
 
     WOLFPROV_ENTER(WP_LOG_COMP_DH, "wp_dh_gen_set_params");
 
     p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_FFC_PBITS);
-    if ((p != NULL) && (!OSSL_PARAM_get_int(p, &ctx->bits))) {
-        ok = 0;
+    if (p != NULL) {
+        if (!OSSL_PARAM_get_int(p, &bits)) {
+            ok = 0;
+        }
+        /* Reject a prime size below the provider minimum. */
+        if (ok && (bits < WP_DH_MIN_BITS)) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_KEY_SIZE_TOO_SMALL);
+            ok = 0;
+        }
+        /* Only store the value once it has passed the size check. */
+        if (ok) {
+            ctx->bits = bits;
+        }
     }
     if (ok) {
         p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_DH_PRIV_LEN);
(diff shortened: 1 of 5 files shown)


src/wp_dh_kmgmt.c
```diff
@@ -1584,8 +1584,14 @@ static int wp_dh_gen_set_params(wp_DhGenCtx* ctx, const OSSL_PARAM params[])
...
-    if ((p != NULL) && (!OSSL_PARAM_get_int(p, &ctx->bits))) {
-        ok = 0;
+    if (p != NULL) {
+        if (!OSSL_PARAM_get_int(p, &ctx->bits)) {
+            ok = 0;
+        }
+        /* Reject a prime size below the provider minimum. */
+        if (ok && (ctx->bits < WP_DH_MIN_BITS)) {
+            ok = 0;
+        }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Regression test does not exercise relocated generation-time validation** — `test/test_dh.c:1355-1368`

*Review generated by Skoll.*

test/test_dh.c
```diff
@@ -1331,4 +1331,67 @@ int test_dh_fromdata_oversize(void *data)
...
+    int err = 0;
+    int rc;
+    EVP_PKEY_CTX *ctx = NULL;
+    EVP_PKEY *keyParams = NULL;
+
+    (void)data;
+
+    PRINT_MSG("Testing DH parameter generation minimum prime size enforcement");
+
+    /* Below-minimum size must be rejected at set-params time. */
```
aidangarske, line 1568: **🟡 [Medium] Regression test does not exercise relocated generation-time validation**
*💡 SUGGEST*  `test`

The PR moves `wp_dh_params_validate(dh)` so generated or copied parameters are validated in every selection path, but the new weak-size test stops at `EVP_PKEY_CTX_set_dh_paramgen_prime_len()`. With the fail-fast guard in `wp_dh_gen_set_params()` present, this test would still pass if the relocated validation in `wp_dh_gen()` were accidentally reverted. That leaves the main behavioral change under-tested, especially for the caller pattern described in the PR where a caller ignores the set-params return and continues to parameter generation.

**Recommendation:** Keep the existing fail-fast assertion, but add a second assertion covering generation-time rejection so the test fails if the `wp_dh_gen()` validation move is removed. Add a FIPS-specific check that deliberately continues after the below-minimum `set_dh_paramgen_prime_len()` failure and asserts `EVP_PKEY_paramgen()` still fails, exercising the relocated `wp_dh_params_validate()` path for 1024-bit DH in FIPS.

yosuke-wolfssl: Thanks, I reworked the regression test.

aidangarske (dismissed):
Looks good!

src/wp_dh_kmgmt.c
```diff
@@ -1897,12 +1898,13 @@ static wp_Dh* wp_dh_gen(wp_DhGenCtx *ctx, OSSL_CALLBACK *cb, void *cbArg)
         else if (!wp_dh_gen_copy_parameters(ctx, dh)) {
             ok = 0;
         }
+        /* Validate parameters meet minimum requirements in every path. */
+        if (ok && (!wp_dh_params_validate(dh))) {
```
ColtonWilley, line 1938: My only worry is we now do a DH params validate on all cases, not only OSSL_KEYMGMT_SELECT_KEYPAIR right

src/wp_dh_kmgmt.c
```diff
@@ -1621,8 +1615,15 @@ static int wp_dh_gen_set_params(wp_DhGenCtx* ctx, const OSSL_PARAM params[])
     WOLFPROV_ENTER(WP_LOG_COMP_DH, "wp_dh_gen_set_params");
 
     p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_FFC_PBITS);
-    if ((p != NULL) && (!OSSL_PARAM_get_int(p, &ctx->bits))) {
-        ok = 0;
+    if (p != NULL) {
+        if (!OSSL_PARAM_get_int(p, &ctx->bits)) {
```
padelsbach: We should not update `ctx->bits` until we've validated it. Maybe use a local in this call, then assign after the size check.

yosuke-wolfssl: You are correct, I fixed it as suggested.

yosuke-wolfssl:
Hello @padelsbach ,
Could you please review this again ?
Thank you in advance.

padelsbach (approved):

PR #428, opened 2026-07-03
Report FIPS module failures through wolfssl_prov_is_running()

yosuke-wolfssl:
## Summary

Fixes finding 4540: in FIPS builds, `wolfssl_prov_is_running()` unconditionally
returned 1, so a wolfCrypt integrity/POST failure was invisible to callers polling
`OSSL_PROV_PARAM_STATUS`. This gates the provider's running/status report on the live
FIPS module status and adds a regression test for the new behavior.

## Problem

- `wolfssl_prov_is_running()` had only one non-1 path — the `WP_CHECK_FORCE_FAIL` debug
  lever — and otherwise always returned 1.
- `wolfprov_get_params()` sets `OSSL_PROV_PARAM_STATUS` directly from
  `wolfssl_prov_is_running()`, so a status-polling caller always saw "running".
- The `wp_fipsCb` FIPS callback only logged and mutated no state, so a failure it
  received was dropped.

Net effect: a FIPS-aware caller enforcing a fail-stop policy via `OSSL_PROV_PARAM_STATUS`
could not observe an unhealthy FIPS module. (The wolfCrypt primitives themselves still
return errors in `FIPS_MODE_FAILED`, so this is a **status-reporting / policy fail-open**,
not silent fail-open crypto.)

## Fix

Gate `wolfssl_prov_is_running()` on `wolfCrypt_GetStatus_fips()` under `#ifdef HAVE_FIPS`;
a non-zero status (integrity/POST or continuous-test failure) returns 0. Querying live
status is a single source of truth and does not depend on the callback having fired.
`wp_fipsCb` is left as a logger.

```c
#ifdef HAVE_FIPS
    if (wolfCrypt_GetStatus_fips() != 0) {
        WOLFPROV_LEAVE(WP_LOG_COMP_PROVIDER, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), 0);
        return 0;
    }
#endif
```

## Regression test

New `test/standalone/tests/fips_status/` standalone test. It loads `libwolfprov`, then:

- **Healthy:** asserts `OSSL_PROV_PARAM_STATUS == 1` **and** that an AES-256-CBC init
  succeeds (proves the provider is actually operable, not just reporting healthy).
- **After a forced FIPS failure:** asserts status flips to `0` **and** the same operation
  is now rejected.

The failure is injected by a compile-time mechanism ladder chosen from what the build
provides:

1. `HAVE_FIPS && HAVE_FORCE_FIPS_FAILURE` → real `wolfCrypt_SetStatus_fips(IN_CORE_FIPS_E)`.
2. `HAVE_FIPS && __linux__` → ELF symbol interposition of `wolfCrypt_GetStatus_fips()`,
   guarded by a `dlsym(RTLD_DEFAULT, ...)` canary that fails loudly if interposition is
   not actually in effect (never misreports an ineffective inject as a provider regression).
3. otherwise → healthy path only, self-reported.

`HAVE_FORCE_FIPS_FAILURE` is enabled by an **internal, test-only** env knob
(`WOLFPROV_FORCE_FIPS_FAILURE`) — deliberately not a CLI flag, so it is not exposed to
customers. When set, `-DHAVE_FORCE_FIPS_FAILURE` reaches both the wolfSSL FIPS build and
the provider/test builds automatically via wolfSSL's generated `options.h` (no relay
variable or source-tree grep).

## CI

`fips-ready.yml` (the source-FIPS job on `ubuntu-22.04`) builds with the knob, runs the
test, and greps for the forced-failure marker so the job **fails if the negative branch
did not actually execute** (guards against silent degradation).

## Files

- `src/wp_wolfprov.c` — the fix; move the FIPS header include to the top.
- `test/standalone/tests/fips_status/{test_fips_status.c,run.sh}` — new test + driver.
- `test/standalone/include.am`, `test/standalone/runners/run_standalone_tests.sh` —
  register the test.
- `scripts/utils-wolfssl.sh` — internal `WOLFPROV_FORCE_FIPS_FAILURE` knob.
- `.github/workflows/fips-ready.yml` — build knob + test step + negative-path assertion.

## Testing

- Both failure-injection branches (`SetStatus` and interposition) compile against the
  real FIPS headers.
- Master standalone runner: all pass.

### DIFF
--- a/.github/workflows/fips-ready.yml
+++ b/.github/workflows/fips-ready.yml
@@ -57,6 +57,10 @@ jobs:
           variant: fips-ready
           openssl_ref: ${{ matrix.openssl_ref }}
           wolfssl_fixed_key: wolfssl-fips-${{ matrix.wolfssl_bundle_ref }}-${{ matrix.openssl_ref }}
+          # The build below sets WOLFPROV_FIPS_FORCE_FAIL, which adds
+          # -DHAVE_FORCE_FIPS_FAILURE to the wolfSSL build; key the cached
+          # install on it so a wolfSSL built without the define is never reused.
+          extra_key: forcefail
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Download FIPS Ready Bundle
@@ -80,7 +84,10 @@ jobs:
 
       - name: Build wolfProvider with FIPS Ready Bundle
         run: |
+          # WOLFPROV_FIPS_FORCE_FAIL is an internal, test-only knob (no CLI
+          # flag) that enables HAVE_FORCE_FIPS_FAILURE for the fips_status test.
           OPENSSL_TAG=${{ matrix.openssl_ref }} \
+          WOLFPROV_FIPS_FORCE_FAIL=1 \
           ./scripts/build-wolfprovider.sh --fips-bundle="$FIPS_BUNDLE_PATH" \
           --fips-check=ready --wolfssl-ver=v${{matrix.wolfssl_bundle_ref}}-stable
 
@@ -104,3 +111,12 @@ jobs:
 
           # --- force-fail mode ---
           WOLFPROV_FORCE_FAIL=1 ./scripts/cmd_test/do-cmd-tests.sh
+
+      - name: Run FIPS Status Test
+        run: |
+          # Exercises the FIPS-status gate in wolfssl_prov_is_running(). In a
+          # FIPS build the test fails unless the forced-failure branch actually
+          # ran, so the exit status is the whole check.
+          export WOLFSSL_ISFIPS=1
+          source scripts/env-setup
+          ./test/standalone/tests/fips_status/run.sh
--- a/test/standalone/include.am
+++ b/test/standalone/include.am
@@ -17,8 +17,8 @@ noinst_HEADERS += test/standalone/test_common.h \
 # Standalone test programs
 # Each test compiles to its own binary for isolated execution
 # Note: These are NOT in check_PROGRAMS because they must be run through scripts, not directly
-noinst_PROGRAMS += test/sha256_simple.test test/hardload.test test/fips_baseline.test test/pqc_interop.test
-DISTCLEANFILES += test/.libs/sha256_simple.test test/.libs/hardload.test test/.libs/fips_baseline.test test/.libs/pqc_interop.test
+noinst_PROGRAMS += test/sha256_simple.test test/hardload.test test/fips_baseline.test test/pqc_interop.test test/fips_status.test
+DISTCLEANFILES += test/.libs/sha256_simple.test test/.libs/hardload.test test/.libs/fips_baseline.test test/.libs/pqc_interop.test test/.libs/fips_status.test
 
 # Common flags for all standalone tests
 STANDALONE_COMMON_CPPFLAGS = -DCERTS_DIR='"$(abs_top_srcdir)/certs"' \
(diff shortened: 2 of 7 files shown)


test/standalone/tests/fips_status/test_fips_status.c
```diff
@@ -0,0 +1,243 @@
...
+{
+#ifdef WP_USE_INTERPOSITION
+    /* Canary: default lookup must resolve to our definition, else the
+     * provider is not seeing the interposed status - fail loudly rather
+     * than misreport an ineffective injection as a provider regression. */
+    if (dlsym(RTLD_DEFAULT, "wolfCrypt_GetStatus_fips") !=
+            (void*)wolfCrypt_GetStatus_fips) {
+        TEST_ERROR("Symbol interposition not in effect");
+        return 1;
+    }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

test/standalone/include.am
```diff
@@ -60,6 +60,10 @@ test_pqc_interop_test_CPPFLAGS = $(STANDALONE_COMMON_CPPFLAGS)
 test_pqc_interop_test_SOURCES = test/standalone/tests/pqc_interop/test_pqc_interop.c
 test_pqc_interop_test_LDADD = $(STANDALONE_COMMON_LDADD)
 
+test_fips_status_test_CPPFLAGS = $(STANDALONE_COMMON_CPPFLAGS)
+test_fips_status_test_SOURCES = test/standalone/tests/fips_status/test_fips_status.c
+test_fips_status_test_LDADD = $(STANDALONE_COMMON_LDADD)
```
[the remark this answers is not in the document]
yosuke-wolfssl: Fixed

.github/workflows/fips-ready.yml
```diff
@@ -82,3 +85,14 @@ jobs:
...
+      - name: Run FIPS Status Test
+        run: |
+          # Exercises the FIPS-status gate in wolfssl_prov_is_running():
+          export WOLFSSL_ISFIPS=1
+          source scripts/env-setup
+          set -o pipefail
+          ./test/standalone/tests/fips_status/run.sh 2>&1 | tee fips-status.log
+          # The forced-failure branch must have actually run in this FIPS build;
+          # fail if the test silently degraded to the healthy path only.
+          grep -q "correctly rejected after FIPS failure" fips-status.log
```
padelsbach: Is there a more generic and thus more robust check than looking for this specific string? Maybe look for a pass/fail?

.github/workflows/fips-ready.yml
```diff
@@ -68,7 +68,10 @@ jobs:
 
       - name: Build wolfProvider with FIPS Ready Bundle
         run: |
-          ./scripts/build-wolfprovider.sh --fips-bundle="$FIPS_BUNDLE_PATH" \
+          # WOLFPROV_FORCE_FIPS_FAILURE is an internal, test-only knob (no CLI
```
padelsbach: Can we call this `WOLFPROV_FIPS_FORCE_FAIL` for consistency with `WOLFPROV_FORCE_FAIL`?

padelsbach (changes requested):

.github/workflows/fips-ready.yml
yosuke-wolfssl: Renamed to WOLFPROV_FIPS_FORCE_FAIL — agreed, it reads better next to WOLFPROV_FORCE_FAIL. Only two references (the CI step and scripts/utils-wolfssl.sh), both updated. The compile-time define it enables (HAVE_FORCE_FIPS_FAILURE) is a wolfSSL-side name, so that one stays as-is.

.github/workflows/fips-ready.yml
yosuke-wolfssl: Thanks for review, and I replaced the grep entirely; CI now just runs the test and checks the exit status.

padelsbach (approved):

PR #430, opened 2026-07-08
Constrain file store decoders to the caller's provider boundary

yosuke-wolfssl:
# Constrain file store decoders to the caller's provider boundary

## Summary

The wolfProvider file store fetched its terminal decoders using only the
structure property and ignored the caller's property query, so a store load
routed to wolfProvider could add and use a decoder from another provider. This
PR propagates the caller's property query into those explicit fetches so a
`provider=wolfprov` boundary is honored consistently. Fixes finding f_5534.

## Problem

`wp_file_set_decoder()` fetched each decoder like this:

```c
decoder = OSSL_DECODER_fetch(ctx->provCtx->libCtx, wp_decoders[i].name,
    wp_decoders[i].propQuery);   /* e.g. "structure=PrivateKeyInfo" */
```

The caller-supplied `ctx->propQuery` (where an application expresses a
`provider=wolfprov` compliance boundary via `OSSL_STORE_PARAM_PROPERTIES`) was
applied only to `OSSL_DECODER_CTX_add_extra`, not to these explicit fetches.
As a result the store's own terminal decoders were fetched without any provider
constraint and could resolve to another provider's implementation.

## Fix

Add a file-local helper that combines each decoder's structure query with the
caller's property query, and use it for every explicit fetch:

```c
static int wp_file_decoder_prop_query(const char* base, const char* caller,
    char** query);   /* 1 = ok, 0 = alloc failure; *query freed by caller */
```

- No caller query (`NULL`) → behavior unchanged (`"structure=..."`).
- `provider=wolfprov` → combined query (`"structure=...,provider=wolfprov"`),
  so the boundary applies to the explicit fetches as well as to
  `OSSL_DECODER_CTX_add_extra`.

## Behavior / compatibility

- Default behavior (no property query) is unchanged.
- When the caller sets `provider=wolfprov`, the boundary is now enforced on all
  of the store's decoder fetches. A load that previously resolved silently
  through another provider will now stay within wolfProvider — or fail if
  wolfProvider cannot satisfy it, which is the intended boundary semantics.

## Testing

- Builds clean under the project's strict warning set
  (`-Werror -Wextra -Wshadow -Wshorten-64-to-32 ...`).
- Full unit suite passes (`OPENSSL_CONF=provider.conf`), including the
  store-load paths: `test_rsa_load_key`, `test_rsa_load_cert`,
  `test_ec_load_key`, `test_ec_load_cert`.

### DIFF
--- a/src/wp_file_store.c
+++ b/src/wp_file_store.c
@@ -32,6 +32,8 @@
 #include <wolfprovider/settings.h>
 #include <wolfprovider/alg_funcs.h>
 
+#include <wolfssl/wolfcrypt/types.h>
+
 /* TODO: support directory access. */
 
 /**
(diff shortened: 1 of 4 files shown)


src/wp_file_store.c
```diff
@@ -366,19 +417,28 @@ static int wp_file_set_decoder(wp_FileCtx* ctx, OSSL_DECODER_CTX* decCtx)
...
     OSSL_DECODER* decoder;
+    char* query = NULL;
 
     WOLFPROV_ENTER(WP_LOG_COMP_PROVIDER, "wp_file_set_decoder");
 
     for (i = 0; ok && (i < WP_DECODERS_SIZE); i++) {
-        decoder = OSSL_DECODER_fetch(ctx->provCtx->libCtx, wp_decoders[i].name,
-            wp_decoders[i].propQuery);
-        if (decoder == NULL) {
+        if (!wp_file_decoder_prop_query(wp_decoders[i].propQuery,
```
aidangarske, line 425: **🟡 [Medium] New property-bound decoder fetch path lacks regression coverage**
*💡 SUGGEST*  `test`

The PR's core behavior is that explicit `OSSL_DECODER_fetch()` calls now include `ctx->propQuery`, but the existing store-load tests do not pass `OSSL_STORE_PARAM_PROPERTIES`; repository search found that parameter only in `src/wp_file_store.c`. That leaves the fixed provider-boundary path untested, so a future change could drop the caller query again while the current RSA/EC load tests still pass.

**Recommendation:** Add a store-load regression test that passes `OSSL_STORE_PARAM_PROPERTIES` with `provider=wolfprov` and verifies the key/cert load succeeds within wolfProvider. A negative case using a property query that excludes wolfProvider (e.g. `provider=default`) would also exercise the failure boundary. Add these focused tests for `OSSL_STORE_PARAM_PROPERTIES` on the file store path before or soon after merge.

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **New property-bound decoder fetch path lacks regression coverage** — `src/wp_file_store.c:425-431`

*Review generated by Skoll.*

src/wp_file_store.c
yosuke-wolfssl: Agreed, and added the regression tests

yosuke-wolfssl:
Hello @aidangarske ,
Could you check this please ?

aidangarske (approved):

PR #435, opened 2026-07-09
fix: guard RSA refCnt decrement with held lock

MarkAtwood:
## Bug

In `wp_rsa_free()` (src/wp_rsa_kmgmt.c) the reference-count decrement `cnt = --rsa->refCnt;` runs unconditionally, outside the `if (rc == 0)` mutex-lock-success guard. If `wc_LockMutex()` fails, two threads can race the unlocked decrement, both observe `cnt == 0`, and double-free the object.

This is an asymmetry with the sibling `wp_rsa_up_ref()` in the same file, which correctly guards its increment behind a successfully-held lock.

## Fix

Move the decrement inside the lock-success branch and default `cnt = 1`. On the rare lock failure the object is leaked rather than double-freed (leak-safe), and in the normal path the decrement is atomic with the held lock.

## Verification

- Compiled the patched translation unit (default multi-threaded build; `--enable-singlethreaded` defaults to no, so `WP_SINGLE_THREADED` is undefined): `gcc -c src/wp_rsa_kmgmt.c` -> exit 0, no errors.
- Preprocessed output confirms the compiled body now performs the decrement only inside `if (rc == 0)`.

Reported by static analysis (Fenrir finding 515). Severity is low: wolfSSL's pthread `wc_LockMutex` essentially never fails for a valid normal-type mutex, so this is a latent defect.

`[fenrir-sweep:held]`

### DIFF
--- a/src/wp_rsa_kmgmt.c
+++ b/src/wp_rsa_kmgmt.c
@@ -525,13 +525,13 @@ static wp_Rsa* wp_rsa_base_new(WOLFPROV_CTX* provCtx, int type)
 void wp_rsa_free(wp_Rsa* rsa)
 {
     if (rsa != NULL) {
-        int cnt;
+        int cnt = 1; /* non-zero default: on lock failure do not free (leak-safe) */
     #ifndef WP_SINGLE_THREADED
         int rc;
 
         rc = wc_LockMutex(&rsa->mutex);
-        cnt = --rsa->refCnt;
         if (rc == 0) {
+            cnt = --rsa->refCnt;
             wc_UnLockMutex(&rsa->mutex);
         }
     #else


src/wp_rsa_kmgmt.c
```diff
@@ -525,13 +525,13 @@ static wp_Rsa* wp_rsa_base_new(WOLFPROV_CTX* provCtx, int type)
 void wp_rsa_free(wp_Rsa* rsa)
 {
     if (rsa != NULL) {
-        int cnt;
+        int cnt = 1; /* non-zero default: on lock failure do not free (leak-safe) */
```
aidangarske, line 528: **🟡 [Medium] Lock-failure refcount behavior lacks regression coverage**
*💡 SUGGEST*  `test`

The PR changes the multi-threaded lock-failure path so `wp_rsa_free()` leaves `cnt` nonzero and skips both the unlocked decrement and object destruction. That is the intended safer behavior, but I found no test or fault-injection coverage for the changed branch. The PR verification notes only compilation and preprocessing, so a future refactor could move `--rsa->refCnt` back outside the successful-lock guard without a regression test catching it.

**Suggestion:**
```suggestion
Add a focused fault-injection or mocked-mutex test that forces `wc_LockMutex(&rsa->mutex)` to fail and verifies `wp_rsa_free()` does not decrement `refCnt` or free the RSA object on that path.
```

aidangarske (commented):
## 🐺 Skoll Code Review

### Posted findings
- [Medium] **Lock-failure refcount behavior lacks regression coverage** — `src/wp_rsa_kmgmt.c:528-534`

*Review generated by Skoll.*

MarkAtwood:
Closing: superseded by #441 (a3f50b7), which moved the refCnt decrement in wp_rsa_free under the lock and keeps the object on lock failure.

PR #439, opened 2026-07-10
Prevent overflow in wp_dh_derive and wp_dh_derive_secret

padelsbach:
While investigating #419, found a missing check in `wp_dh_derive`. This change adds a small check of the buffer size and test case.

### DIFF
--- a/src/wp_dh_exch.c
+++ b/src/wp_dh_exch.c
@@ -298,13 +298,19 @@ static int wp_dh_derive_secret(wp_DhCtx* ctx, unsigned char* secret,
         }
         else {
             /* Front pad with zeros if required. */
-            if (ctx->pad && (len != maxLen)) {
+            if (ctx->pad && ((size_t)len < maxLen)) {
                 XMEMMOVE(secret + maxLen - len, secret, len);
                 XMEMSET(secret, 0, maxLen - len);
                 len = (word32)maxLen;
             }
-            /* Return length of data in buffer. */
-            *secLen = len;
+            else if ((size_t)len > maxLen) {
+                ok = 0;
+            }
+
+            if (ok) {
+                /* Return length of data in buffer. */
+                *secLen = len;
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/wp_dh_exch.c
```diff
@@ -356,9 +362,14 @@ static int wp_dh_derive(wp_DhCtx* ctx, unsigned char* secret,
         maxLen = wp_dh_get_size(ctx->key);
 
         if (ctx->kdfType == WP_KDF_NONE) {
```
aidangarske, line 364: **🟠 [High] Unpadded DH derive now rejects valid shorter outputs**
*🚫 BLOCK*  `bug`

The new no-KDF size check rejects every caller buffer smaller than the DH prime length before deriving. That changes the unpadded DH contract: with `OSSL_EXCHANGE_PARAM_PAD` disabled, DH output is not padded and may legitimately be shorter than the prime. The fixed test vector used by this PR has a 256-byte prime but a 255-byte unpadded shared secret, so a 255-byte buffer can hold the actual output. Before this PR that case could succeed; after this PR it always fails because `secSize < maxLen`. This effectively makes unpadded derive require a padded-size buffer while still returning an unpadded length.

**Recommendation:** For the no-KDF/no-pad path, avoid writing `wc_DhAgree()` directly into an undersized caller buffer, but do not reject solely because `secSize < maxLen`. Derive into a temporary `maxLen` buffer when needed, then copy the actual unpadded result if `outLen <= secSize`. Keep the immediate `secSize < maxLen` rejection only for padded output, where the actual output length is guaranteed to be `maxLen`. Update the regression test to cover a shorter-than-prime unpadded secret that fits in the caller buffer.

aidangarske (changes requested):
## 🐺 Skoll Code Review

### Posted findings
- [High] **Unpadded DH derive now rejects valid shorter outputs** — `src/wp_dh_exch.c:364-372`

*Review generated by Skoll.*

src/wp_dh_exch.c
padelsbach: We're updating the contract for two reasons:

1. The size of the unpadded secret is not known by the caller -- depends on the leading zeros when generating it. Best to just require the full max size.
2. The new check mirrors OpenSSLs behavior. See `if (outlen < dhsize)` in `providers/implementations/exchange/dh_exch.c`

I'll push an update which A/B tests WP vs OSSL to confirm equiv behavior

aidangarske (approved):
Looks good

padelsbach:
jenkins retest this please

ColtonWilley (approved):

PR #440, opened 2026-07-10
wp_seed_src: read /dev/urandom via raw fd to avoid seccomp SIGSYS

ColtonWilley:
## Problem

SEED-SRC read `/dev/urandom` through a buffered stdio `FILE*` (`XFOPEN`). That breaks OpenSSH's privsep preauth child:

1. `sshd` primes its RNG (`seed_rng` -> `RAND_status`) before the privsep fork, so the first `fread` pulls a block of read-ahead into the stdio buffer.
2. The child inherits the buffered `FILE*` and its fd.
3. On child exit, glibc stdio cleanup calls `lseek()` to discard the unread read-ahead.
4. `lseek()` is not in OpenSSH's preauth seccomp allow-list, so the child is killed with `SIGSYS` (audit shows `sig=31 syscall=8`).

Stock OpenSSL reads the device with a raw fd and never hits this; the buffered `FILE*` was the defect. Separately, re-initializing SEED-SRC (a second provider load) orphaned the previous buffered handle.

## Fix

- Read `/dev/urandom` with a raw fd (`open`/`read`/`close`, `O_CLOEXEC`). No stdio buffer, so no read-ahead and no exit-time `lseek`.
- Share the fd and the wolfSSL seed callback across provider contexts, reference counted. The last unload closes the fd and restores the seed callback to `wc_GenerateSeed` — not `NULL`, which would make a later `wc_InitRng` fail `DRBG_NO_SEED_CB` under FIPS.
- SEED-SRC now depends on wolfSSL file support; `NO_WOLFSSL_DIR` builds `#error` rather than silently falling back.

## Tests

- `test_seccomp_sandbox` — runs under the verbatim OpenSSH 9.9p1 preauth filter: single-context (the original crash), re-init, and multi-context survivor. Compiled in with `-DWP_TEST_SECCOMP_SANDBOX` (enabled by the `seed-src` CI job; a skip stub otherwise, since it needs `PR_SET_SECCOMP`).
- `test_seed_src_refcount` — multi-context refcount; the first unload must not close the shared fd.
- `test_seed_src_reload` — drives the refcount to zero (fd close + seed-callback restore) and reloads; re-exec'd so the count actually reaches zero, which the main suite's own reference otherwise prevents.

### DIFF
--- a/src/wp_seed_src.c
+++ b/src/wp_seed_src.c
@@ -70,10 +70,8 @@
  * seccomp sandbox that blocks open()/openat(). */
 static int g_urandom_fd = XBADFD;
 
-/* Identity of the device behind g_urandom_fd, captured by fstat at open time.
- * A host application (e.g. OpenSSH's privsep setup) may dup2 its own descriptor
- * onto our cached fd number; before reading we re-fstat and confirm the fd still
- * refers to the same /dev/urandom, reopening if it was reused. */
+/* Device identity of g_urandom_fd, captured by fstat at open, so a later read
+ * can detect when another subsystem has reused (dup2'd over) our fd number. */
 static int    g_urandom_id_valid = 0;
 static dev_t  g_urandom_dev;
 static ino_t  g_urandom_ino;
@@ -108,10 +106,6 @@ static int wp_urandom_open_locked(void)
             return -1;
         }
 
-        /* Record the device identity so a later read can detect fd reuse. Uses
-         * a raw fstat(): wolfSSL exposes XSTAT (path-based) but no fd-based
-         * XFSTAT, and stat-ing the path cannot tell us our fd was reused. This
-         * matches the raw open() above (there is likewise no XOPEN). */
         {
             struct stat st;
             if (fstat(g_urandom_fd, &st) == 0) {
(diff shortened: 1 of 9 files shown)


src/wp_seed_src.c
```diff
@@ -60,115 +63,161 @@
...
+ * @return  -1 on failure.
+ */
+#ifdef WC_RNG_SEED_CB
+static int wp_wolfssl_seed_cb(OS_Seed* os, byte* seed, word32 sz)
+{
+    int rc;
+
+    (void)os;
+
+    if (WP_URANDOM_LOCK() != 0) {
```
padelsbach, line 230: In addition to the lock, should probably check that the refcount >= 1 just to make sure the file is still open

src/wp_seed_src.c
```diff
@@ -60,115 +63,161 @@
...
- * This callback is registered with wc_SetSeed_Cb() and is called by wolfSSL's
- * DRBG when it needs entropy (including after fork detection).
+ * @return  0 on success.
+ * @return  -1 if /dev/urandom cannot be opened.
+ */
+static int wp_urandom_open_locked(void)
+{
+    if (g_urandom_fd == XBADFD) {
+        do {
+            g_urandom_fd = open(URANDOM_PATH, O_RDONLY | O_CLOEXEC);
```
padelsbach, line 100: Can we use `XOPEN` here for consistency?

padelsbach (commented):
Good improvement. Couple nits

src/wp_seed_src.c
ColtonWilley: There is no XOPEN :)

src/wp_seed_src.c
ColtonWilley: Good catch, changes made

padelsbach (approved):

PR #445, opened 2026-07-14
Free FP ECC cache in provider teardown

ColtonWilley:
`wc_ecc_free()` doesn't release wolfCrypt's fixed-point ECC cache (`fp_cache[]`), and nothing in wolfProvider calls `wc_ecc_fp_free()` — so it's held for the process lifetime. Call it in `wolfssl_prov_ctx_free()`, guarded by `FP_ECC`.

Frees the teardown thread's cache (thread-local under `HAVE_THREAD_LS`); no-op when empty.


ColtonWilley:
Jenkins retest this please.

PR #447, opened 2026-07-15
Fix ECX/ECC get_params to size export buffer from data_size

yosuke-wolfssl:
## Summary

Fixes finding f_2236: ECX and ECC `get_params` used `p->return_size` instead of
`p->data_size` as the export buffer size.

Per the `OSSL_PARAM` contract, `data_size` is the caller's buffer capacity while
`return_size` is the provider's output field for the actual data length. Three
`get_params` helpers seeded the wolfSSL export length (`outLen`, an in/out
parameter) from `return_size`. When a param is reused — e.g. it carries a stale
or zero `return_size` from a prior size-query call — the export function receives
a wrong (often 0) buffer size and truncates or fails. It only worked for
freshly-constructed params because OpenSSL initializes `return_size` to an
unmodified sentinel rather than a small value.

## Changes

- `src/wp_ecx_kmgmt.c`
  - `wp_ecx_get_params_enc_pub_key`: `outLen` now seeded from `p->data_size`.
  - `wp_ecx_get_params_priv_key`: same fix.
- `src/wp_ecc_kmgmt.c`
  - `wp_ecc_get_params_enc_pub_key`: same fix.
- `test/test_ecx.c` (+ `test/unit.c`, `test/unit.h`)
  - New `test_ecx_x25519_get_params_stale_ret`: constructs params with a valid
    32-byte buffer but forces `return_size = 0` (the reused-param condition),
    then calls `EVP_PKEY_get_params` for both the private and public key and
    verifies the returned bytes (against the RFC 7748 §6.1 Alice key pair) and
    `return_size`.

The `p->return_size = outLen` output assignments at the end of each helper are
unchanged — those are correct. A repo-wide sweep of `return_size`/`data_size`
uses confirmed no other helper (ML-DSA, ML-KEM, ML-X, DH, RSA) misused
`return_size` as an input; those already size from `data_size`.

## Testing

- New test passes; reverting the fix makes it fail (`priv key get_params failed
  with stale return_size`), confirming it catches the bug.
- All 34 ECX/ECC/ECDH/eckeygen unit tests pass.

### DIFF
--- a/src/wp_ecc_kmgmt.c
+++ b/src/wp_ecc_kmgmt.c
@@ -747,7 +747,7 @@ static int wp_ecc_get_params_enc_pub_key(wp_Ecc* ecc, OSSL_PARAM params[],
     p = OSSL_PARAM_locate(params, key);
     if (p != NULL) {
         int rc;
-        word32 outLen = (word32)p->return_size;
+        word32 outLen = (word32)p->data_size;
 
         if (ecc->hasPub == 0) {
             ok = 0;
--- a/test/test_ecx.c
+++ b/test/test_ecx.c
@@ -758,6 +758,109 @@ int test_ecx_x25519_raw_priv_roundtrip(void *data)
     return err;
 }
 
+/* get_params must size the export buffer from the caller's data_size, not
+ * return_size. A reused param can carry a stale/zero return_size from a prior
+ * size query; using it would truncate or fail the export. */
+int test_ecx_x25519_get_params_stale_ret(void *data)
+{
+    int err = 0;
+    EVP_PKEY *pkey = NULL;
+    static const unsigned char privKey[] = {
+        0x77, 0x07, 0x6d, 0x0a, 0x73, 0x18, 0xa5, 0x7d,
+        0x3c, 0x16, 0xc1, 0x72, 0x51, 0xb2, 0x66, 0x45,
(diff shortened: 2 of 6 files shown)


test/test_ecx.c
```diff
@@ -758,6 +758,86 @@ int test_ecx_x25519_raw_priv_roundtrip(void *data)
     return err;
 }
 
+/* get_params must size the export buffer from the caller's data_size, not
+ * return_size. A reused param can carry a stale/zero return_size from a prior
+ * size query; using it would truncate or fail the export. */
+int test_ecx_x25519_get_params_stale_ret(void *data)
```
yosuke-wolfssl, line 764: Fixed

src/wp_ecc_kmgmt.c
```diff
@@ -747,7 +747,7 @@ static int wp_ecc_get_params_enc_pub_key(wp_Ecc* ecc, OSSL_PARAM params[],
     p = OSSL_PARAM_locate(params, key);
     if (p != NULL) {
         int rc;
-        word32 outLen = (word32)p->return_size;
+        word32 outLen = (word32)p->data_size;
```
yosuke-wolfssl, line 750: Fixed

SparkiDev (approved):

PR #453, opened 2026-07-27
Encrypt PKCS#8 keys when a cipher is set on the PrivateKeyInfo encoder

aidangarske:
`openssl pkey -aes256` silently wrote unencrypted private keys for `EC, Ed25519, X25519, DH and ML-DSA` (RSA was fine) because the encoders only encrypted when the caller explicitly asked for the `EncryptedPrivateKeyInfo` structure, ignoring a cipher set on the ordinary `PrivateKeyInfo` path now they encrypt whenever a cipher is set plus a DH fix so generated keys don't encode an empty private value.

CLI-testing the new SLH-DSA encoders using `openssl pkey -aes256` emitted `BEGIN PRIVATE KEY` instead of `BEGIN ENCRYPTED PRIVATE KEY`

Added PKCS8 openssl scripts testing

### DIFF
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -69,6 +69,8 @@ struct wp_Dh {
 #ifndef WP_SINGLE_THREADED
     /** Mutex for reference count updating. */
     wolfSSL_Mutex mutex;
+    /** Mutex for key data access and synchronization. */
+    wolfSSL_Mutex keyMutex;
 #endif
     /** Count of references to this object. */
     int refCnt;
@@ -414,6 +416,16 @@ static wp_Dh* wp_dh_new(WOLFPROV_CTX *provCtx)
                 ok = 0;
             }
         }
+        if (ok) {
+            rc = wc_InitMutex(&dh->keyMutex);
+            if (rc != 0) {
+                WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG,
+                    "wc_InitMutex", rc);
+                wc_FreeMutex(&dh->mutex);
+                wc_FreeDhKey(&dh->key);
+                ok = 0;
+            }
+        }
     #endif
         if (ok) {
             dh->refCnt = 1;
@@ -461,6 +473,7 @@ void wp_dh_free(wp_Dh* dh)
             OPENSSL_free(dh->pub);
             OPENSSL_clear_free(dh->priv, dh->privSz);
     #ifndef WP_SINGLE_THREADED
+            wc_FreeMutex(&dh->keyMutex);
             wc_FreeMutex(&dh->mutex);
     #endif
             wc_FreeDhKey(&dh->key);
@@ -2626,10 +2639,6 @@ static int wp_dh_encode_spki(const wp_Dh *dh, unsigned char* keyData,
     }
     if (ok) {
         *keyLen = len;
-        /* wolfSSL calculating it wrong. */
-        if (keyData[1] == 0x81) {
-            keyData[2] = len - 3;
-        }
     }
 
     WOLFPROV_LEAVE(WP_LOG_COMP_DH, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
@@ -2638,12 +2647,13 @@ static int wp_dh_encode_spki(const wp_Dh *dh, unsigned char* keyData,
 
 /**
  * Copy a generated private key into the inner wolfSSL key if not already set.
+ * The caller must hold dh->keyMutex.
  *
  * @param [in]  dh  DH key object.
  * @return  1 on success.
  * @return  0 on failure.
  */
-static int wp_dh_sync_priv_to_key(const wp_Dh *dh)
+static int wp_dh_sync_priv_to_key(const wp_Dh* dh)
 {
     int ok = 1;
     int ret;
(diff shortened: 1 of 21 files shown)


src/wp_dh_kmgmt.c
```diff
@@ -2684,6 +2706,14 @@ static int wp_dh_encode_pki_size(const wp_Dh *dh, size_t* keyLen)
  * @return  1 on success.
  * @return  0 on failure.
  */
+/* wolfSSL calculating it wrong. */
```
padelsbach: Looks like this function got inserted between `wp_dh_encode_pki` and it's doxygen block above

src/wp_internal.c
```diff
@@ -925,9 +925,18 @@ int wp_cipher_from_params(const OSSL_PARAM params[], int* cipher,
...
+            if (cipherName != NULL) {
+                *cipherName = NULL;
+            }
+        }
+        else {
+            size_t i = WP_CIPHER_NAMES_LEN;
 
+#ifdef WP_HAVE_PKCS8_ENC
             for (i = 0; i < WP_CIPHER_NAMES_LEN; i++) {
                 if ((XSTRLEN(wp_cipher_names[i].name) == p->data_size) &&
```
padelsbach, line 949: AI says `wp_cipher_names` should be wrapped in `#ifdef WP_HAVE_PKCS8_ENC` to avoid an unused var here

src/wp_internal.c
padelsbach: Should we remove existing uses of `#if defined(HAVE_PKCS8) && !defined(NO_PWDBASED)` in favor of `#if WP_HAVE_PKCS8_ENC`?

test/test_pkey.c
```diff
@@ -298,6 +298,128 @@ int test_pkey_dec(EVP_PKEY *pkey, OSSL_LIB_CTX* libCtx, unsigned char *msg,
...
+    OSSL_ENCODER_CTX* ectx = NULL;
+    OSSL_DECODER_CTX* dctx = NULL;
+    OSSL_DECODER_CTX* bctx = NULL;
+    unsigned char* data = NULL;
+    size_t dataLen = 0;
+    size_t encLen = 0;
+    const unsigned char* pp;
+    const char* pass = "wolfprov-test-pass";
+    const char* badPass = "wrong-passphrase";
+    size_t passLen = strlen(pass);
```
padelsbach: Nit: there are a couple new uses of plain `strlen` that should be `XSTRLEN`

src/wp_internal.c
```diff
@@ -925,9 +925,18 @@ int wp_cipher_from_params(const OSSL_PARAM params[], int* cipher,
         if (p->data_type != OSSL_PARAM_UTF8_STRING) {
             ok = 0;
         }
-        if (ok) {
-            size_t i;
+        else if (p->data == NULL) {
```
padelsbach, line 936: AI says this path is not covered by new tests

src/wp_ecc_kmgmt.c
```diff
@@ -2931,6 +2931,15 @@ static int wp_ecc_encode(wp_EccEncDecCtx* ctx, OSSL_CORE_BIO *cBio,
         ok = 0;
     }
 
+    /* Traditional PEM encryption is not implemented, so refuse a cipher here
+     * rather than write the private key in the clear. */
+    if (ok && ((ctx->format == WP_ENC_FORMAT_TYPE_SPECIFIC) ||
+               (ctx->format == WP_ENC_FORMAT_X9_62)) &&
+            (ctx->cipherName != NULL) &&
+            ((selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) != 0)) {
+        ok = 0;
```
padelsbach, line 2942: AI says there might not be a test case for this

test/test_ecc.c
```diff
@@ -992,6 +992,22 @@ int test_ecc_encode_epki(void *data)
     }
 
     EVP_PKEY_free(osslKey);
+    if (err == 0) {
```
padelsbach, line 1026: General comment about testing: do you think we could add test cases to the shell scripts for black-box testing? Eg in `ecc-cmd-test.sh`, etc

src/wp_dh_kmgmt.c
aidangarske: Fixed in commit 2f6e94a. Moved wp_dh_fix_pki_len() before the associated wp_dh_encode_pki() Doxygen block so the declaration and documentation stay together.

test/test_ecc.c
aidangarske: This is a useful general testing improvement, but it is outside the SLH-DSA/PKCS#8 fix scope. I left it as follow-up work rather than expanding this dependency PR.

test/test_pkey.c
aidangarske: Fixed

src/wp_internal.c
aidangarske: Fixed

src/wp_internal.c
aidangarske: Wrapped wp_cipher_names and its length macro with #ifdef WP_HAVE_PKCS8_ENC

src/wp_internal.c
aidangarske: tested now

src/wp_ecc_kmgmt.c
aidangarske: tested now

padelsbach (dismissed):
Would prefer if we had an actual command line test as shown in the PR description, but can be a follow up

aidangarske:
> Would prefer if we had an actual command line test as shown in the PR description, but can be a follow up

I agree let me add that

padelsbach (dismissed):

padelsbach:
LGTM, let's give @ColtonWilley a chance to look at this one

ColtonWilley (approved):

PR #454, opened 2026-07-28
F-4936 - Take parent DRBG lock around raw dispatch calls

ColtonWilley:
Child DRBGs called the parent's get_seed/clear_seed via raw dispatch pointers with no lock held, so concurrent child instantiations (e.g. two threads' first RAND_bytes) raced on the shared primary's WC_RNG.

Fix (mirrors OpenSSL's own PROV_DRBG):
- capture parent ENABLE_LOCKING/LOCK/UNLOCK in newctx
- take the parent lock around parentGetSeed/parentClearSeed
- propagate enable_locking to the parent, one-shot; publish the local mutex only after init
- scrub the seed in place if the clear_seed lock fails

F-5186 reviewed with this: works as designed — EVP_RAND locking is opt-in and core serializes generate/uninstantiate once enabled. Full suite green, default and --enable-seed-src.

### DIFF
--- a/src/wp_drbg.c
+++ b/src/wp_drbg.c
@@ -62,6 +62,14 @@ typedef struct wp_DrbgCtx {
     OSSL_FUNC_rand_get_seed_fn* parentGetSeed;
     /** Parent's clear_seed function. */
     OSSL_FUNC_rand_clear_seed_fn* parentClearSeed;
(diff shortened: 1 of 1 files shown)


ColtonWilley:
Jenkins retest this please.

PR #458, opened 2026-08-03
Test ML-KEM mixed-NULL KEM requests

aidangarske:
The provider already implements OpenSSL-compatible mixed-NULL handling in e44edc66.

Add EVP_PKEY encapsulate/decapsulate regression coverage to ensure mixed-NULL requests are rejected while valid size queries remain supported.

### DIFF
--- a/test/test_mlkem.c
+++ b/test/test_mlkem.c
@@ -53,12 +53,13 @@ static const mlkem_test_level mlkem_levels[] = {
  * @param [out] pkey  Generated EVP_PKEY (caller frees).
  * @return  0 on success, non-zero on failure.
  */
-static int wp_test_mlkem_keygen(const char* name, EVP_PKEY** pkey)
+static int wp_test_mlkem_keygen_ex(OSSL_LIB_CTX* libCtx, const char* prop,
+    const char* name, EVP_PKEY** pkey)
 {
     int err = 0;
     EVP_PKEY_CTX* ctx = NULL;
 
-    ctx = EVP_PKEY_CTX_new_from_name(wpLibCtx, name, NULL);
+    ctx = EVP_PKEY_CTX_new_from_name(libCtx, name, prop);
     err = (ctx == NULL);
     if (err == 0) {
         err = EVP_PKEY_keygen_init(ctx) != 1;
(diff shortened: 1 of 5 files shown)


test/test_mlkem.c
```diff
@@ -333,6 +339,83 @@ int test_mlkem_encap_decap(void* data)
...
+    size_t secretLen = sizeof(secret);
+
+    err = wp_test_mlkem_keygen_ex(libCtx, prop, level->name, &pkey);
+    if (err == 0) {
+        ectx = EVP_PKEY_CTX_new_from_pkey(libCtx, pkey, prop);
+        err = (ectx == NULL) ||
+            (EVP_PKEY_encapsulate_init(ectx, NULL) != 1);
+    }
+    if (err == 0) {
+        results[0] = EVP_PKEY_encapsulate(ectx, NULL, &ctLen, secret,
```
padelsbach: Can you add checks/compares of `ctLen` and `secretLen` for the all calls?

test/test_mlkem.c
```diff
@@ -333,6 +339,83 @@ int test_mlkem_encap_decap(void* data)
...
+        results[1] = EVP_PKEY_encapsulate(ectx, ct, &ctLen, NULL,
+            &secretLen);
+    }
+    if (err == 0) {
+        dctx = EVP_PKEY_CTX_new_from_pkey(libCtx, pkey, prop);
+        err = (dctx == NULL) ||
+            (EVP_PKEY_decapsulate_init(dctx, NULL) != 1);
+    }
+    if (err == 0) {
+        results[2] = EVP_PKEY_decapsulate(dctx, secret, &secretLen, NULL,
```
padelsbach: Let's reset `secretLen` here prior to this call as done at line 364

padelsbach (changes requested):

test/test_mlkem.c
aidangarske: done

test/test_mlkem.c
aidangarske: done

padelsbach (commented):
looks good, will merge when CI wraps up

padelsbach (approved):

PR #460, opened 2026-08-06
Validate DH peer public key subgroup before key agreement

aidangarske:
```
F-5914
```


test/test_dh.c
```diff
@@ -1735,4 +1736,486 @@ int test_dh_pgen_min_bits(void *data)
...
+    if (err == 0) {
+        ctx = EVP_PKEY_CTX_new_from_pkey(wpLibCtx, key, NULL);
+        err = ctx == NULL;
+    }
+    if (err == 0) {
+        err = EVP_PKEY_derive_init(ctx) != 1;
+    }
+    if (err == 0) {
+        rc = EVP_PKEY_derive_set_peer(ctx, small);
+        if (rc == 1) {
```
padelsbach, line 2128: Why do we only test if `set_peer` succeeds?

src/wp_dh_exch.c
```diff
@@ -708,4 +773,3 @@ const OSSL_DISPATCH wp_dh_keyexch_functions[] = {
 };
 
 #endif /* WP_HAVE_DH */
-
```
padelsbach, line 711: Nit: the empty trailing line is dropped. Can we keep it?

padelsbach (changes requested):

PR #461, opened 2026-08-06
F-4287: check public point is on curve

padelsbach:
`wp_ecdh_set_peer` now runs a full public key check (on-curve + correct order). `wp_ecdh_derive_secret` re-checks the point before use.

Relatedly, fixes the quick-validate path, which was dead code behind USE_ECC_B_PARAM and had an inverted wc_ecc_point_is_on_curve test plus a curve id passed where an index was expected.

Fixes F-4287


padelsbach:
jenkins retest this please

PR #463, opened 2026-08-07
Resolve the wolfSSL FIPS Ready bundle at CI run time

aidangarske:
# Description

- FIPS Ready CI was pinned to wolfssl-5.8.2, hardcoded in fips-ready.yml and perf-regression.yml.
--enable-fips=ready is not a fixed FIPS version: it moved from v7 to v8 as of bundle 5.9.2.
- CI kept testing FIPS v7 while the actual FIPS Ready download had moved to v8 with no coverage.
- wolfProvider did not build against FIPS v8: SHA-512/224 and SHA-512/256 were registered in the
provider dispatch table but their definitions stayed gated on HAVE_FIPS, so v8 modules hit an
undefined-reference link error.
- New scripts/fetch-fips-ready.sh resolves, downloads, and verifies FIPS Ready bundles at CI run
time instead of a hardcoded version, with bounded retry and checksum verification.
_discover-versions.yml gains fips_ready_latest/_pr_array/_all_array outputs; PR runs test the
back-compat anchors plus newest, nightly sweeps every hosted bundle.
- Fixed the v8 SHA-512/224/256 registration bug by routing both sides through the same
WP_HAVE_SHA512_224/256 capability macros; added RSA dupctx tests exercising the affected
wp_hash_copy() branches.
- Discovery failures fail loudly (new fips_ready_resolved gate) instead of silently testing a
one-version floor and reporting green.


.github/workflows/_discover-versions.yml
```diff
@@ -169,11 +194,61 @@ jobs:
...
+          FR_LATEST=$(echo "$FR_ALL" | tail -n 1)
+
+          # `ready` does not name a fixed FIPS module; map each hosted bundle
+          # to the module it actually builds (from wolfSSL's configure.ac
+          # `ready|vN-ready` case, checked per tag). Extend the boundaries
+          # below when a new FIPS generation ships.
+          fips_module_for() {
+            local key
+            key=$(printf '%s' "$1" | awk -F. '{ printf "%05d.%05d.%05d", $1, $2, $3 }')
+            if   [[ "$key" > "00005.00009.00001" ]]; then echo "v8"  # > 5.9.1
```
padelsbach: is the fixed mapping still in place with this code? would it make sense to pull it from `HAVE_FIPS_VERSION_MAJOR`?

.github/workflows/_discover-versions.yml
```diff
@@ -169,11 +194,61 @@ jobs:
           OSSL_ALL_JSON=$(printf '%s\n' "$OSSL_ALL" | jq -R . | jq -s -c .)
           OSSL_LATEST=$(echo "$OSSL_ALL" | tail -n 1)
 
+          # FIPS Ready: every bundle hosted on wolfssl.com at or above the floor.
+          # Floor 5.8.2: the oldest bundle wolfProvider is tested against.
+          FR_FLOOR="5.8.2"
+          FR_OK=true
+          # Bound the probe so a hung wolfssl.com cannot eat this shared job's
+          # timeout and take all ~40 callers down; a timeout degrades to FR_OK=false.
+          FR_ALL=$(timeout 240 ./scripts/fetch-fips-ready.sh --list --floor "$FR_FLOOR") || true
```
padelsbach: can we limit the script run to fips-ready, or do we need this script to run on _all_ workflows?

scripts/fetch-fips-ready.sh
```diff
@@ -0,0 +1,318 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Resolves and downloads wolfSSL FIPS Ready bundles from wolfssl.com.
```
padelsbach, line 4: Shell script for scraping the web feels fragile and probably the wrong tool for the job, esp since we are relying upon it for the SHA's and integrity. 

Could we do one of the following instead?
1. commit a text file with the versions and wait for a nightly failure to update
2. better yet, roll the bundles into the ghcr image
4. other ideas?

ColtonWilley (dismissed):

.github/workflows/nightly-fips-ready.yml
```diff
@@ -0,0 +1,196 @@
...
+          }
+          for ver in $HOSTED; do
+            if oras manifest fetch "$REPO:$ver" >/dev/null 2>&1; then
+              echo "already in ghcr: $ver"
+              continue
+            fi
+            echo "publishing new bundle: $ver"
+            # fetch-fips-ready.sh verifies the download before it returns.
+            DIR=$(./scripts/fetch-fips-ready.sh "$ver" --dest "$GITHUB_WORKSPACE")
+            ./scripts/oras-build-deps.sh push "$REPO:$ver" "$DIR"
```
padelsbach, line 77: Why keep the fallback? 

I think we need to check the output here.

aidangarske: checking output and removed fallback now

.github/workflows/nightly-fips-ready.yml
```diff
@@ -0,0 +1,196 @@
...
+    timeout-minutes: 45
+    steps:
+      - name: Checkout wolfProvider
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - name: Install ORAS
+        run: |
+          set -euo pipefail
+          ORAS_VERSION="1.2.2"
```
padelsbach: There are now 3 instances of ORAS installation. Can we unify into a script?

.github/workflows/fips-ready.yml
```diff
@@ -91,6 +110,20 @@ jobs:
...
+          # Authoritative source: the FIPS version wolfSSL's own build resolved
+          # `ready` to, read from HAVE_FIPS_VERSION_MAJOR rather than guessed
+          # from the bundle version.
+          OPTS="wolfssl-install/include/wolfssl/options.h"
+          if [ -f "$OPTS" ]; then
+            MAJ=$(grep -oE '#define +HAVE_FIPS_VERSION_MAJOR +[0-9]+' "$OPTS" | awk '{print $3}')
+            echo "Bundle ${{ matrix.wolfssl_bundle_ref }} builds FIPS module v${MAJ:-unknown}"
+            echo "FIPS module v${MAJ:-unknown} from bundle ${{ matrix.wolfssl_bundle_ref }}" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "::warning::wolfSSL options.h not found at $OPTS; cannot read HAVE_FIPS_VERSION_MAJOR"
```
padelsbach: should we fail the job when this occurs?

ColtonWilley (approved):

PR #464, opened 2026-08-07
Add Windows/MSVC port

ColtonWilley:
Adds a Visual Studio solution under IDE/WINVS and the source changes needed to build wolfProvider with MSVC, FIPS and non-FIPS. Four configurations, x64.

user_settings.h is hand-written and committed rather than generated. options.h can't be reused here: it's an autotools output that bakes in results that are wrong under MSVC. wolfSSL and wolfProvider have to be built against the same one.

include/wolfprovider/version.h is now tracked. It was gitignored, so a fresh clone couldn't build on Windows at all.

Built and tested FIPS and non-FIPS, in the VS IDE.


include/wolfprovider/version.h
```diff
@@ -0,0 +1,39 @@
+/* version.h.in
+ *
+ * Copyright (C) 2006-2025 wolfSSL Inc.
```
padelsbach, line 3: Rather than check this in, can we put this in `IDE/WINVS`?

If not, AI says we need to update configure.ac, remove version.hin, etc

src/wp_logging.c
```diff
@@ -468,18 +468,18 @@ static void wolfProv_LogLevelToMask(const char* level, size_t len, void* ctx) {
...
         uint32_t    mask;
     } log_levels[] = {
-        { "WP_LOG_LEVEL_ERROR",     XSTRLEN("WP_LOG_LEVEL_ERROR"),  WP_LOG_LEVEL_ERROR   },
-        { "WP_LOG_LEVEL_ENTER",     XSTRLEN("WP_LOG_LEVEL_ENTER"),  WP_LOG_LEVEL_ENTER   },
-        { "WP_LOG_LEVEL_LEAVE",     XSTRLEN("WP_LOG_LEVEL_LEAVE"),  WP_LOG_LEVEL_LEAVE   },
-        { "WP_LOG_LEVEL_INFO",      XSTRLEN("WP_LOG_LEVEL_INFO"),   WP_LOG_LEVEL_INFO    },
-        { "WP_LOG_LEVEL_VERBOSE",   XSTRLEN("WP_LOG_LEVEL_VERBOSE"),WP_LOG_LEVEL_VERBOSE },
-        { "WP_LOG_LEVEL_DEBUG",     XSTRLEN("WP_LOG_LEVEL_DEBUG"),  WP_LOG_LEVEL_DEBUG   },
-        { "WP_LOG_LEVEL_TRACE",     XSTRLEN("WP_LOG_LEVEL_TRACE"),  WP_LOG_LEVEL_TRACE   },
+        { "WP_LOG_LEVEL_ERROR",     sizeof("WP_LOG_LEVEL_ERROR") - 1,  WP_LOG_LEVEL_ERROR   },
```
padelsbach, line 471: nit: Could be nice to macro-ize this for DRYness and compactness with a macro to define each entry

include/wolfprovider/internal.h
```diff
@@ -199,6 +201,8 @@ void wp_provctx_unlock_rng(WOLFPROV_CTX* provCtx);
 #define WP_CAST_ALGO_DH     6
 #define WP_CAST_ALGO_COUNT  7
 
+int wp_init_cast_mutexes(void);
```
padelsbach, line 204: Should we apply the same preprocessor bounds as in the definition: `#if defined(HAVE_FIPS) && !defined(WP_SINGLE_THREADED)`

test/test_ecx.c
```diff
@@ -821,6 +821,14 @@ int test_ecx_x25519_raw_priv_roundtrip(void *data)
  * size query; using it would truncate or fail the export. */
 int test_ecx_x25519_get_params_stale_ret(void *data)
 {
+#ifdef WOLFSSL_CURVE25519_BLINDING
+    /* Blinded curve25519 needs an RNG attached with wc_curve25519_set_rng(),
+     * which wolfProvider does not do yet. Gated on the macro rather than
+     * _WIN32 so builds without blinding keep the coverage. */
+    PRINT_MSG("X25519 get_params test skipped - curve25519 blinding enabled");
+    (void)data;
```
padelsbach, line 829: Nit: the return before the declarations is not C99. Could add the `#else` below

include/wolfprovider/version.h
ColtonWilley: I dont think thats a good idea actually, that setup is practically begging for a version.h clash if include paths are off even a little. AI is just overly upset, wolfSSL proper uses this exact same setup.

test/test_ecc.c
```diff
@@ -1364,6 +1364,14 @@ static const unsigned char x25519_shared_secret[] = {
  */
 int test_ecdh_x25519_vector(void *data)
 {
+#ifdef WOLFSSL_CURVE25519_BLINDING
+    /* Blinded curve25519 needs an RNG attached with wc_curve25519_set_rng(),
+     * which wolfProvider does not do yet. Gated on the macro rather than
+     * _WIN32 so builds without blinding keep the coverage. */
+    PRINT_MSG("X25519 ECDH KAT skipped - curve25519 blinding enabled");
+    (void)data;
+    return 0;
```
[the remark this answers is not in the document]
ColtonWilley: This is intentional, curve25519 blinding breaks on linux as well with same build config. Proper X25519 RNG handling for blinding will be a follow up PR.

include/wolfprovider/version.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef WOLFPROV_VERSION_H
+#define WOLFPROV_VERSION_H
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define LIBWOLFPROV_VERSION_STRING "1.2.1"
+#define LIBWOLFPROV_VERSION_HEX 0x01002001
```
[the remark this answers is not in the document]
ColtonWilley: version.h structure is intentional and mirrors wolfssl proper

include/wolfprovider/internal.h
ColtonWilley: Its just the proto, implementation is bounded properly already

padelsbach (approved):

test/test_ecx.c
ColtonWilley: Agreed, it has been reworked

PR #465, opened 2026-08-11
Fenrir fixes in wp_kbkdf.c

padelsbach:
- **F-8681: check result of param lookup before comparison**
- **F-2789: add MODE, SEED, and INFO to KBKDF settable params**
- **F-2243: zeroize hmac/cmac context when freeing**

### DIFF
--- a/src/wp_kbkdf.c
+++ b/src/wp_kbkdf.c
@@ -177,15 +177,18 @@ static int wp_kdf_kbkdf_set_ctx_params(wp_KbkdfCtx* ctx,
                 if (!OSSL_PARAM_get_utf8_string_ptr(p, &mode)) {
                     ok = 0;
                 }
-                if (XSTRCMP(mode, "COUNTER") == 0) {
-                    ctx->mode = WP_KDF_MODE_COUNTER;
-                }
-                else if (XSTRCMP(mode, "FEEDBACK") == 0) {
-                    ctx->mode = WP_KDF_MODE_FEEDBACK;
-                }
-                else {
-                    WOLFPROV_MSG(WP_LOG_COMP_KDF, "Invalid KDF mode: %s", mode);
-                    ok = 0;
+                if (ok) {
+                    if (XSTRCMP(mode, "COUNTER") == 0) {
+                        ctx->mode = WP_KDF_MODE_COUNTER;
+                    }
+                    else if (XSTRCMP(mode, "FEEDBACK") == 0) {
+                        ctx->mode = WP_KDF_MODE_FEEDBACK;
+                    }
+                    else {
+                        WOLFPROV_MSG(WP_LOG_COMP_KDF, "Invalid KDF mode: %s",
+                            mode);
+                        ok = 0;
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/wp_kbkdf.c
```diff
@@ -526,6 +541,11 @@ static int wp_kbkdf_mac_update(wp_KbkdfCtx* ctx, const unsigned char *data,
 static void wp_kbkdf_mac_free(wp_KbkdfCtx* ctx)
```
aidangarske, line 543: **Cleanup can dispatch through the wrong MAC type**

`macInited` records only a boolean, while cleanup selects the object using mutable `ctx->mac`. If a derive initializes HMAC and then update or final fails, `macInited` remains set. A subsequent set-params call can change `ctx->mac` to CMAC before reset or free. The newly added cleanup then calls `wc_CmacFree()` on the uninitialized CMAC object instead of releasing the live HMAC, leaking its allocations or device state and invoking the wrong cleanup routine. The reverse transition has the same problem. This incorrect cross-type cleanup is introduced by the new boolean bookkeeping and clear-time call.

**Fix:** Release any initialized MAC on every derive exit before parameters can be changed, or record the initialized MAC type separately and dispatch cleanup using that immutable value.

test/test_kbkdf.c
```diff
@@ -377,6 +377,210 @@ static int test_kbkdf_counter(void)
...
+    }
+    *p++ = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_KEY, key,
+        sizeof(key));
+    *p++ = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_SALT, label,
+        sizeof(label));
+    *p++ = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_INFO, context,
+        sizeof(context));
+    *p = OSSL_PARAM_construct_end();
+
+    if (EVP_KDF_derive(kctx, first, sizeof(first), params) <= 0) {
```
aidangarske, line 554: **Reset test never resets an initialized MAC**

The first derive succeeds, and each successful derive iteration already calls `wp_kbkdf_mac_free()`, leaving `macInited` equal to zero before `EVP_KDF_CTX_reset()`. Consequently, this test exercises only the new guard's no-op path and would also pass before the F-2243 cleanup change. It cannot detect missing or incorrect reset-time cleanup of a live HMAC or CMAC.

**Fix:** Add fault injection that fails MAC update or final after successful initialization, then reset or free the context. Cover both MAC types and switching from HMAC to CMAC and vice versa.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [High] **Cleanup can dispatch through the wrong MAC type** — `src/wp_kbkdf.c:541-564`
- [Medium] **Reset test never resets an initialized MAC** — `test/test_kbkdf.c:554-560`

src/wp_kbkdf.c
padelsbach: fixed, using tag rather than boolean

test/test_kbkdf.c
padelsbach: added test cases

aidangarske (approved):

aidangarske:
Jenkins retest this please

padelsbach:
jenkins retest this please

PR #467, opened 2026-08-13
Validate EC_PUB_X/EC_PUB_Y on import by routine them through the X9.63 point import

gasbytes:
Route EC_PUB_x/EC_PUB_Y through wc_ecc_import_x963_ex, the same validated import the encoded public key path uses, so the point is checked to be onn the curve,
both ordinates are required together, and the projective z ordinate is set.

Also, added some 3 regression tests for the off-curve and X-without-Y cases to test the edge cases of this behaviour.

Fixes F-4694.

### DIFF
--- a/src/wp_ecc_kmgmt.c
+++ b/src/wp_ecc_kmgmt.c
@@ -537,13 +537,99 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
     static const OSSL_PARAM wp_ecc_supported_settable_params[] = {
         OSSL_PARAM_int(OSSL_PKEY_PARAM_USE_COFACTOR_ECDH, NULL),
         OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY, NULL, 0),
-        OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_EC_PUB_X, NULL, 0),
-        OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_EC_PUB_Y, NULL, 0),
+        OSSL_PARAM_BN(OSSL_PKEY_PARAM_EC_PUB_X, NULL, 0),
+        OSSL_PARAM_BN(OSSL_PKEY_PARAM_EC_PUB_Y, NULL, 0),
         OSSL_PARAM_int(OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC, NULL),
         OSSL_PARAM_END
     };
     (void)provCtx;
     return wp_ecc_supported_settable_params;
+}
+
+/**
+ * Import an X9.63 encoded public key point into the ECC key object.
+ *
+ * @param [in, out] ecc   ECC key object.
+ * @param [in]      data  X9.63 encoded public key point.
+ * @param [in]      len   Length of encoded point in bytes.
+ * @return  1 on success.
+ * @return  0 on failure.
+ */
+static int wp_ecc_import_pub_x963(wp_Ecc* ecc, const unsigned char* data,
+    word32 len)
+{
+    int ok = 1;
+    int rc;
+    int init = 0;
+    ecc_key pub;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_ECC, "wp_ecc_import_pub_x963");
+
+    rc = wc_ecc_init_ex(&pub, NULL, INVALID_DEVID);
+    if (rc != 0) {
+        WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_init_ex", rc);
+        ok = 0;
+    }
+    else {
+        init = 1;
+    }
+    if (ok) {
+        rc = wc_ecc_import_x963_ex(data, len, &pub, ecc->curveId);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG,
+                "wc_ecc_import_x963_ex", rc);
+            ok = 0;
+        }
+    }
+    /* Curve may have been resolved from the point size when no curve was set
+     * on the key object yet. */
+    if (ok && (pub.dp == NULL)) {
+        ok = 0;
+    }
+    if (ok) {
+        rc = wc_ecc_check_key(&pub);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_check_key",
+                rc);
+            ok = 0;
+        }
+    }
+    if (ok) {
+        rc = wc_ecc_set_curve(&ecc->key, 0, pub.dp->id);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_set_curve",
+                rc);
+            ecc->hasPub = 0;
+            ok = 0;
+        }
+    }
+    if (ok) {
+        rc = wc_ecc_copy_point(&pub.pubkey, &ecc->key.pubkey);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_copy_point",
+                rc);
+            ecc->hasPub = 0;
+            ok = 0;
+        }
+    }
+    if (ok) {
+        /* Only the public part was set - keep any private key in place. */
+        if ((ecc->key.type == ECC_PRIVATEKEY) ||
+            (ecc->key.type == ECC_PRIVATEKEY_ONLY)) {
+            ecc->key.type = ECC_PRIVATEKEY;
+        }
+        else {
+            ecc->key.type = ECC_PUBLICKEY;
+        }
+        ecc->hasPub = 1;
+    }
+
+    if (init) {
+        wc_ecc_free(&pub);
+    }
+
+    WOLFPROV_LEAVE(WP_LOG_COMP_ECC, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
+    return ok;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


test/test_ecc.c
```diff
@@ -2873,6 +2873,230 @@ static int test_ec_import_group_no_nul(void)
...
+
+    if (err == 0) {
+        err = (dctx = EVP_PKEY_CTX_new_from_pkey(wpLibCtx, priv,
+                NULL)) == NULL;
+    }
+    if (err == 0) {
+        err = EVP_PKEY_derive_init(dctx) != 1;
+    }
+    if (err == 0 && EVP_PKEY_derive_set_peer_ex(dctx, peer, 0) > 0 &&
+            EVP_PKEY_derive(dctx, secret, &secretLen) > 0) {
```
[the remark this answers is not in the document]
gasbytes: the test itself is testing p-256, so the derive length is always 32 bytes, I don't think the query itself adds any meaningfuly covverage here (?) marking and rejecting as nit in my opinion.

src/wp_ecc_kmgmt.c
```diff
@@ -584,35 +584,126 @@ static int wp_ecc_set_params_enc_pub_key(wp_Ecc *ecc, const OSSL_PARAM params[],
...
+
+        rc = wc_ecc_import_x963_ex(point, 1 + (2 * (word32)size), &ecc->key,
+            ecc->curveId);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG,
+                "wc_ecc_import_x963_ex", rc);
             ok = 0;
         }
         if (ok) {
+            /* wc_ecc_import_x963_ex only checks the point against the curve
```
padelsbach: According to the logic in this comment, we should also apply the check below in `wp_ecc_set_params_enc_pub_key`. Maybe add a helper and call from both?

iiuc, `WOLFSSL_VALIDATE_ECC_IMPORT` is fairly new (or at least the checking is), so we want to keep this check for older wolfSSL versions.

src/wp_ecc_kmgmt.c
```diff
@@ -584,35 +584,126 @@ static int wp_ecc_set_params_enc_pub_key(wp_Ecc *ecc, const OSSL_PARAM params[],
...
+    if (ok && setX) {
+        if (mp_to_unsigned_bin_len(&y, point + 1 + size, size) != MP_OKAY) {
+            ok = 0;
+        }
+    }
+    if (ok && setX) {
+        int rc;
+        int origType;
+
+        rc = wc_ecc_import_x963_ex(point, 1 + (2 * (word32)size), &ecc->key,
```
padelsbach: AI tells me this call can overwrite/zero existing fields within the ecc key, which is fine for import, but potentially dangerous when called from `wp_ecc_set_params`. Can you investigate?

src/wp_ecc_kmgmt.c
```diff
@@ -584,35 +584,126 @@ static int wp_ecc_set_params_enc_pub_key(wp_Ecc *ecc, const OSSL_PARAM params[],
...
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG,
+                "wc_ecc_import_x963_ex", rc);
             ok = 0;
         }
         if (ok) {
+            /* wc_ecc_import_x963_ex only checks the point against the curve
+             * when wolfSSL is built with WOLFSSL_VALIDATE_ECC_IMPORT, so check
+             * it here instead of relying on the build options. */
+            origType = ecc->key.type;
```
padelsbach: AI says `ecc->key.type` is always `ECC_PUBLICKEY` so the save+restore is not needed. But zooming out, do we need to check the key we just imported?

src/wp_ecc_kmgmt.c
padelsbach: Not directly related, but stumbled upon this block where X and Y are `octet_string`s. I think they should be `OSSL_PARAM_BN`

padelsbach (changes requested):

src/wp_ecc_kmgmt.c
gasbytes: yup you are right, will make it a helper and will apply the check there too.
thanks!

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,58 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+    word32 len)
+{
+    int ok = 1;
+    int rc;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_ECC, "wp_ecc_import_pub_x963");
+
+    rc = wc_ecc_import_x963_ex(data, len, &ecc->key, ecc->curveId);
+    if (rc != 0) {
+        WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_import_x963_ex",
```
padelsbach: I think we need to clear `hasPub` on this failure path

src/wp_ecc_kmgmt.c
```diff
@@ -568,51 +611,112 @@ static int wp_ecc_set_params_enc_pub_key(wp_Ecc *ecc, const OSSL_PARAM params[],
...
+static int wp_ecc_set_params_pub(wp_Ecc *ecc, const OSSL_PARAM params[])
+{
+    int ok = 1;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_ECC, "wp_ecc_set_params_pub");
+
+    if (!wp_ecc_set_params_pub_xy(ecc, params)) {
         ok = 0;
     }
     if (wp_ecc_set_params_enc_pub_key(ecc, params,
```
padelsbach: For completeness, can you add the `(ok == 1)` condition to the other checks in this function?

test/test_ecc.c
```diff
@@ -2873,6 +2873,230 @@ static int test_ec_import_group_no_nul(void)
...
+        err = (params = OSSL_PARAM_BLD_to_param(bld)) == NULL;
+    }
+    if (err == 0) {
+        err = (ctx = EVP_PKEY_CTX_new_from_name(wpLibCtx, "EC", NULL)) == NULL;
+    }
+    if (err == 0) {
+        err = EVP_PKEY_fromdata_init(ctx) != 1;
+    }
+    if (err == 0) {
+        ret = EVP_PKEY_fromdata(ctx, pkey, EVP_PKEY_PUBLIC_KEY, params) == 1;
```
padelsbach, line 3049: AI tells me we don't have a test which builds an off-curve public key, per the scenario in the Fenrir finding. Can you please check?

Also, is there a positive/valid test case for public import?

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,58 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+ */
+static int wp_ecc_import_pub_x963(wp_Ecc* ecc, const unsigned char* data,
+    word32 len)
+{
+    int ok = 1;
+    int rc;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_ECC, "wp_ecc_import_pub_x963");
+
+    rc = wc_ecc_import_x963_ex(data, len, &ecc->key, ecc->curveId);
```
padelsbach: AI says we should use a local `ecc_key` instead of `ecc->key` here since `wc_ecc_import_x963_ex` can succeed and write the point into the struct, then remain there after a failure below.

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,58 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+    if (rc != 0) {
+        WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_import_x963_ex",
+            rc);
+        ok = 0;
+    }
+    if (ok) {
+        rc = wc_ecc_check_key(&ecc->key);
```
padelsbach: Thinking we should add this check to the other pub key decode calls in this file, `wp_ecc_decode_spki` and `wp_ecc_decode_x963_pub`

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,58 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+ * @return  1 on success.
+ * @return  0 on failure.
+ */
+static int wp_ecc_import_pub_x963(wp_Ecc* ecc, const unsigned char* data,
+    word32 len)
+{
+    int ok = 1;
+    int rc;
+
+    WOLFPROV_ENTER(WP_LOG_COMP_ECC, "wp_ecc_import_pub_x963");
```
padelsbach, line 570: AI says we need a lock around the two steps in this function (import and check). Can you investigate?

padelsbach (changes requested):
More findings, sorry for another big batch

src/wp_ecc_kmgmt.c
gasbytes: yeah it's truthful but not really dangerous in my opinion, since the worst case is an error. the key is marked public-only, so wolfssl refuses to sign or derive with it and master did the same thing here already.
I fixed it anyway since it's cleaner anyway to go through a scratch key, and the private key is left alone instead.

test/test_ecc.c
gasbytes: >AI tells me we don't have a test which builds an off-curve public key, per the scenario in the Fenrir finding. Can you please check?

there is one which is test_ec_import_off_curve_xy(), it takes the valid p-256 point and flips a bit in Y, then feeds it in as x/y.

>Also, is there a positive/valid test case for public import?

adding one that imports valid x/y and checks the result matches and it's correct. thanks.

src/wp_ecc_kmgmt.c
gasbytes: fixed in previous commit, now squashed.

src/wp_ecc_kmgmt.c
gasbytes: fixed in previous commit, now squashed.

src/wp_ecc_kmgmt.c
gasbytes: added the missing checks.

src/wp_ecc_kmgmt.c
gasbytes: looked into it, I don't think that's needed. since both steps operate on pub which is local and there's nothing shared that would cause a race as far as I can tell.

src/wp_ecc_kmgmt.c
```diff
@@ -2185,6 +2336,12 @@ static int wp_ecc_decode_x963_pub(wp_Ecc* ecc, unsigned char* data, word32 len)
     if (rc != 0) {
         ok = 0;
     }
+    if (ok) {
+        rc = wc_ecc_check_key(&ecc->key);
```
padelsbach: `wc_ecc_import_x963` seems to already call `wc_ecc_check_key`

Update: AI says the 3 added calls to `wc_ecc_check_key` can be removed. And `wc_ecc_check_key` includes the check for `WOLFSSL_VALIDATE_ECC_IMPORT`. Can you confirm that?

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,105 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+    }
+    else {
+        init = 1;
+    }
+    if (ok) {
+        rc = wc_ecc_import_x963_ex(data, len, &pub, ecc->curveId);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG,
+                "wc_ecc_import_x963_ex", rc);
+            ok = 0;
```
padelsbach, line 581: should we clear `ecc->hasPub` here, and the other error paths below?

src/wp_ecc_kmgmt.c
```diff
@@ -537,15 +537,105 @@ static const OSSL_PARAM* wp_ecc_settable_params(WOLFPROV_CTX* provCtx)
...
+        rc = wc_ecc_set_curve(&ecc->key, 0, pub.dp->id);
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_ecc_set_curve",
+                rc);
+            ecc->hasPub = 0;
+            ok = 0;
+        }
+    }
+    if (ok) {
+        rc = wc_ecc_copy_point((ecc_point*)&pub.pubkey, &ecc->key.pubkey);
```
padelsbach: Nit: this cast may not be required

src/wp_ecc_kmgmt.c
gasbytes: I don't think we should, since the import writes into the scratch pub key, so on failure ecc still holds whatever valid public key it had. If we clear, hasPub there would discard it. The two below run after wc_ecc_set_curve/wc_ecc_ciopy_point have already modified the key object in ecc (ecc->key), so there the key really is left inconsistent.

src/wp_ecc_kmgmt.c
gasbytes: two can be safely removed, but one needs to stay since wc_ecc_import_x963_ex only calls wc_ecc_check_key when wolfSSL is build with WOLFSSL_CALIDATE_ECC_IMPORT, which CI's --enable-all-crypto doesn't set.

padelsbach (approved):

PR #469, opened 2026-08-17
Cover FIPS-clean EAP methods in hostap CI and enable AES key wrap for wpasupplicant FIPS

aidangarske:
- Needs https://github.com/wolfSSL/osp/pull/357 and https://github.com/wolfSSL/testing/pull/1017 exercises newly patched tests
- Also addresses failing hostap nightly with pin bump


scripts/utils-wolfssl.sh
```diff
@@ -42,7 +42,7 @@ WOLFSSL_SOURCE_DIR=${SCRIPT_DIR}/../wolfssl-source
 WOLFSSL_INSTALL_DIR=${SCRIPT_DIR}/../wolfssl-install
 WOLFSSL_ISFIPS=${WOLFSSL_ISFIPS:-0}
 WOLFSSL_FIPS_VERSION=${WOLFSSL_FIPS_VERSION:-"5.2.4"}
-WOLFSSL_FIPS_CONFIG_OPTS=${WOLFSSL_CONFIG_OPTS:-'--enable-opensslcoexist '}
+WOLFSSL_FIPS_CONFIG_OPTS=${WOLFSSL_CONFIG_OPTS:-'--enable-opensslcoexist --enable-aeskeywrap '}
```
padelsbach: `seed-src.yml` sets `$WOLFSSL_CONFIG_OPTS`, so the coexist and keywrap options won't get applied. Is that ok?

aidangarske: I dont think this is needed since its built directly in jenkins. I removed it

padelsbach (approved):

PR #473, opened 2026-08-26
cmac: take a NULL key in init as a restart with the cached key

yosuke-wolfssl:
### Problem

`wp_cmac_init()` only restarted the wolfSSL CMAC object when a key was supplied, so `EVP_MAC_init(ctx, NULL, 0, NULL)` — the documented way to reset a MAC context while keeping the installed key — did nothing:

| Reset point | Result |
|---|---|
| After `EVP_MAC_final()` | `wc_CmacFinal()` zeroes the `Cmac`, so the next `EVP_MAC_update()` returns 0 |
| Mid-stream, before `final` | the MAC silently covers `msgA‖msgB` instead of `msgB`, reporting success at every call |

Separately, `wp_cmac_update()` and `wp_cmac_final()` accepted a context whose key install had failed. The FIPS bundles' `wc_CmacUpdate()`/`wc_CmacFinal()` validate only pointers — no `cmac->type` check — so a context that never received a usable key returned a MAC computed with an all-zero AES key. Current non-FIPS wolfSSL rejects that via `default: BAD_FUNC_ARG`, which is why it reproduces only under FIPS.

### Fix (`src/wp_cmac.c`)

- **`wp_cmac_set_key()`** takes `key == NULL` as "keep the cached key, just restart": `keyLen` resolves from `macCtx->keyLen`, and a zero resolved length fails. The length and `expKeySize` checks and the cleanse/copy of `macCtx->key` are gated on `key != NULL`, and `macCtx->keyLen` is assigned only after `wc_InitCmac_ex()` succeeds.
- **`wp_cmac_init()`** calls it on every init, so the AES key schedule and the k1/k2 subkeys are re-derived.
- **`wp_cmac_update()` and `wp_cmac_final()`** fail when `macCtx->keyLen` is 0, rather than relying on wolfSSL to reject an uninitialized `Cmac`.

Closes f_11550.

### Tests

`test_cmac_reinit` drives one `EVP_MAC_CTX` through keyless init after `final` and mid-stream, comparing every MAC against OpenSSL, and checks that a keyless restart after a rejected key refuses both `EVP_MAC_update()` and `EVP_MAC_final()`.

### Verification

- Full unit suite green on non-FIPS and against the `v5.2.4` FIPS bundle; clean under `-Werror`.
- Negative controls: reverting `wp_cmac.c` fails the reinit legs; removing only the `final` guard fails the new `final` case.
- Before the guards, a local FIPS build reproduced the `PRB-fips-scripts-test` failure exactly — test 22, `CMAC produced a MAC from a key that was never set`; it passes after.

### DIFF
--- a/src/wp_cmac.c
+++ b/src/wp_cmac.c
@@ -104,7 +104,7 @@ static void wp_cmac_free(wp_CmacCtx* macCtx)
  * Allocates space for the key in the wolfSSL CMAC object.
  *
  * @param [in, out] macCtx   CMAC context object.
- * @param [in]      key      Key data to set.
+ * @param [in]      key      Key data to set. NULL restarts with the cached key.
  * @param [in]      keyLen   Length of key data in bytes.
  * @param [in]      restart  Restart CMAC calculation.
  * @return  1 on success.
@@ -117,33 +117,42 @@ static int wp_cmac_set_key(wp_CmacCtx* macCtx, const unsigned char* key,
 
     WOLFPROV_ENTER(WP_LOG_COMP_MAC, "wp_cmac_set_key");
 
-    if (keyLen > AES_256_KEY_SIZE) {
+    /* A NULL key restarts with the key already cached. */
+    if (key == NULL) {
+        keyLen = macCtx->keyLen;
+    }
+
+    if (keyLen == 0) {
         ok = 0;
     }
-    if (ok && (macCtx->expKeySize != 0) && (keyLen != macCtx->expKeySize)) {
+    if (ok && (key != NULL) && (keyLen > AES_256_KEY_SIZE)) {
         ok = 0;
+    }
+    if (ok && (key != NULL) && (macCtx->expKeySize != 0) &&
+            (keyLen != macCtx->expKeySize)) {
+        ok = 0;
+    }
+    if (ok && (key != NULL)) {
+        /* The cached length stays zero until the key is installed. */
+        OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
+        macCtx->keyLen = 0;
+        XMEMCPY(macCtx->key, key, keyLen);
+    }
+    if (ok && restart) {
+    #if LIBWOLFSSL_VERSION_HEX >= 0x05000000
+        int rc = wc_InitCmac_ex(&macCtx->cmac, macCtx->key, (word32)keyLen,
+            macCtx->type, NULL, NULL, INVALID_DEVID);
+    #else
+        int rc = wc_InitCmac(&macCtx->cmac, macCtx->key, (word32)keyLen,
+            macCtx->type, NULL);
+    #endif
+        if (rc != 0) {
+            WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_InitCmac/wc_InitCmac_ex", rc);
+            ok = 0;
+        }
     }
     if (ok) {
-        if (macCtx->keyLen > 0) {
-            OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
-        }
         macCtx->keyLen = keyLen;
-        XMEMCPY(macCtx->key, key, keyLen);
-
-        if (restart) {
-        #if LIBWOLFSSL_VERSION_HEX >= 0x05000000
-            int rc = wc_InitCmac_ex(&macCtx->cmac, macCtx->key,
-                (word32)macCtx->keyLen, macCtx->type, NULL, NULL,
-                INVALID_DEVID);
-        #else
-            int rc = wc_InitCmac(&macCtx->cmac, macCtx->key,
-                (word32)macCtx->keyLen, macCtx->type, NULL);
-        #endif
-            if (rc != 0) {
-                WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "wc_InitCmac/wc_InitCmac_ex", rc);
-                ok = 0;
-            }
-        }
     }
 
     WOLFPROV_LEAVE(WP_LOG_COMP_MAC, __FILE__ ":" WOLFPROV_STRINGIZE(__LINE__), ok);
@@ -190,7 +199,7 @@ static wp_CmacCtx* wp_cmac_dup(wp_CmacCtx* src)
  * Initializes an CMAC context object with a key and parameters.
  *
  * @param [in, out] macCtx  CMAC context object to initialize.
- * @param [in]      key     Key data to set.
+ * @param [in]      key     Key data to set. NULL restarts with the cached key.
  * @param [in]      keyLen  Length of key data in bytes.
  * @param [in]      params  Extra parameters to set.
  * @return  1 on success.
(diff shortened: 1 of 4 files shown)


src/wp_cmac.c
```diff
@@ -117,18 +117,24 @@ static int wp_cmac_set_key(wp_CmacCtx* macCtx, const unsigned char* key,
...
         ok = 0;
     }
-    if (ok && (macCtx->expKeySize != 0) && (keyLen != macCtx->expKeySize)) {
+    if (ok && (key != NULL) && (keyLen > AES_256_KEY_SIZE)) {
+        ok = 0;
+    }
+    if (ok && (key != NULL) && (macCtx->expKeySize != 0) &&
+            (keyLen != macCtx->expKeySize)) {
         ok = 0;
     }
```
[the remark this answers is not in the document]
yosuke-wolfssl: Thanks — the mechanism is real: with `key == NULL` the `expKeySize` check is skipped. But the cipher parameter is inert on a keyless restart in **both** providers, so the MAC isn't unexpected.

OpenSSL only applies a stored cipher inside `cmac_setkey()`, so with no key the new cipher is recorded but never installed. wolfProvider lands in the same place: the restart passes `macCtx->keyLen` to `wc_InitCmac_ex()`, never `expKeySize`, so the installed key still matches the cipher it was installed with. I ran your sequence both directions:

```
AES-256-CBC + 32B key, final, re-init with AES-128-CBC param, no key:
  openssl  104f51da2c90543a4d782ec5885988f0
  wolfprov 104f51da2c90543a4d782ec5885988f0

AES-128-CBC + 16B key, final, re-init with AES-256-CBC param, no key:
  openssl  e50d766fa637f3440604e998623ce12e
  wolfprov e50d766fa637f3440604e998623ce12e
  (== true AES-128/16B MAC; != true AES-256/32B MAC)
```

Byte-identical, no truncation or padding either way — `wc_InitCmac_ex()` accepts 16/24/32 equally.

So a fail-fast check would make wolfProvider return 0 where OpenSSL returns 1 and a valid MAC — a new divergence rather than a closed gap. `expKeySize` stays correct for its real job: validating the *next* supplied key against the newly requested cipher.

It would only fire on a genuine cipher switch with no new key, so it wouldn't disturb ordinary resets. Happy to add it if maintainers prefer stricter-than-OpenSSL here — just flagging it's a deliberate behavior difference, not a bug fix.

src/wp_cmac.c
```diff
@@ -117,18 +117,24 @@ static int wp_cmac_set_key(wp_CmacCtx* macCtx, const unsigned char* key,
...
 
-    if (keyLen > AES_256_KEY_SIZE) {
+    if ((key == NULL) && (macCtx->keyLen == 0)) {
         ok = 0;
     }
-    if (ok && (macCtx->expKeySize != 0) && (keyLen != macCtx->expKeySize)) {
+    if (ok && (key != NULL) && (keyLen > AES_256_KEY_SIZE)) {
+        ok = 0;
+    }
+    if (ok && (key != NULL) && (macCtx->expKeySize != 0) &&
```
yosuke-wolfssl, line 131: Confirmed and fixed — the second mechanism you name is the one that bites. `wp_cmac_set_key()` assigned `macCtx->keyLen` and copied the key *before* `wc_InitCmac_ex()` ran, so a failed install left a key cached that was never installed, and the `wp_cmac_init()` change in this PR made that stale cache reachable:

```
key set with no cipher (fails), then AES-128-CBC, then EVP_MAC_init(ctx, NULL, 0, NULL):
  openssl      init = 0
  wolfprovider init = 1, mac = 104f51da...   (the AES-256/32B value)
```

Fixed by ordering rather than by the suggested validation: `keyLen` is cleared while the supplied key is copied in, and assigned only once the init succeeds. A rejected key now leaves nothing cached. `test_cmac_reinit_stale_key` covers it against both providers.

I did not add the `expKeySize` check on the NULL-key path — the other route you mention isn't a divergence. On a keyless restart the cipher parameter is inert in both implementations (OpenSSL only applies a stored cipher inside `cmac_setkey()`), and both agree byte for byte in each direction:

```
AES-256 + 32B key, then AES-128 param, no key:  both 104f51da2c90543a4d782ec5885988f0
AES-128 + 16B key, then AES-256 param, no key:  both e50d766fa637f3440604e998623ce12e
```

Validating there would return 0 where OpenSSL returns 1 and a valid MAC.

test/test_cmac.c
```diff
@@ -358,6 +358,142 @@ int test_cmac_multi_update(void *data)
...
+        PRINT_ERR_MSG("CMAC after mid-stream reset covers stale data");
+        err = 1;
+    }
+
+    EVP_MAC_CTX_free(ctx);
+    EVP_MAC_free(emac);
+    return err;
+}
+
+int test_cmac_reinit(void *data)
```
yosuke-wolfssl, line 545: Restored it. That was removed accidentally.

src/wp_cmac.c
```diff
@@ -117,34 +118,43 @@ static int wp_cmac_set_key(wp_CmacCtx* macCtx, const unsigned char* key,
...
-    if (ok && (macCtx->expKeySize != 0) && (keyLen != macCtx->expKeySize)) {
+    if (ok && (key != NULL) && (keyLen > AES_256_KEY_SIZE)) {
         ok = 0;
     }
-    if (ok) {
-        if (macCtx->keyLen > 0) {
-            OPENSSL_cleanse(macCtx->key, macCtx->keyLen);
-        }
-        macCtx->keyLen = keyLen;
+    if (ok && (key != NULL) && (macCtx->expKeySize != 0) &&
```
yosuke-wolfssl, line 131: Accurate description, but it isn't a divergence — OpenSSL behaves identically here, so validating would make wolfProvider stricter than the provider it substitutes for.

On a keyless restart the cipher parameter is inert in both implementations: OpenSSL's `cmac_set_ctx_params()` only applies a stored cipher inside `cmac_setkey()`, so with no key supplied the new cipher is recorded and never installed. Measured on 24f9585, both directions:

```
AES-128-CBC + 16B key, then AES-256-CBC param, no key:
  openssl      e50d766fa637f3440604e998623ce12e   (the AES-128 MAC)
  wolfprovider e50d766fa637f3440604e998623ce12e

AES-256-CBC + 32B key, then AES-128-CBC param, no key:
  openssl      104f51da2c90543a4d782ec5885988f0   (the AES-256 MAC)
  wolfprovider 104f51da2c90543a4d782ec5885988f0
```

Both return 1 in every case. Applying the check on the NULL-key path would return 0 where OpenSSL returns a valid MAC, breaking drop-in behaviour for a request OpenSSL also accepts.

The mismatch is bookkeeping only — `macCtx->size` stays `AES_BLOCK_SIZE`, `macCtx->type` stays `WC_CMAC_AES`, and `expKeySize` still validates the next *supplied* key against the newly requested cipher.

The other mechanism from the previous round — a failed key-set leaving `macCtx->keyLen` committed — was a genuine divergence and is fixed in this commit; `test_cmac_reinit_stale_key` covers it.

Happy to add the strict check if maintainers would rather reject here, but flagging it as a deliberate behaviour difference from OpenSSL rather than a bug fix.

yosuke-wolfssl:
Jenkins retest please

PR #476, opened 2026-08-26
Fix AES-CTS one-block, split-init IV, and TLS record HMAC handling

gasbytes:
* In wp_aes_cts_encrypt/wp_aes_cts_decrypt handle an input of exactly one block as plain CBC;
* In wp_aes_stream_init pass ctx->iv to wc_AesSetKey, which is the cached iv from the wolfprovider context;
* In wp_aes_stream_init restore oiv into iv when re-initializing without an iv, matching wp_aes_block_init;
* In wp_hmac_set_ctx_params consume OSSL_MAC_PARAM_TLS_DATA_SIZE and advertise it in wp_hmac_settable_ctx_params;
* In wp_hmac_final hash dummy blocks so the number of blocks hashed depends only on the padded record length;

Added associated regression test for each change (test_aes128_cts_one_block, test_aes128_cts_split_init and test_hmac_tls_data_size).


src/wp_hmac.c
```diff
@@ -258,6 +263,64 @@ static int wp_hmac_init(wp_HmacCtx* macCtx, const unsigned char* key,
...
+        (macCtx->tlsDataSize < dataLen + macCtx->size + 1)) {
+        return;
+    }
+
+    blockSz = (word32)blockSizeRet;
+    blockMask = blockSz - 1;
+    while (((word32)1 << blockBits) < blockSz) {
+        blockBits++;
+    }
+    padSz = (blockSz >> 3) + 1;
```
aidangarske, line 313: **Use the length-field width when counting final hash blocks**

`padSz` includes the `0x80` byte, causing `wp_hmac_blocks()` to count a padding spill one byte too early. For SHA-256, `realSz` values 55 and 56 produce the same helper count, although wolfCrypt finalization processes one block for 55 bytes and two for 56. Consequently, records with 42 and 43 data bytes can receive the same dummy count at a fixed padded-record size while their total compression counts differ by one. This defeats the equalization introduced by the PR at every padding boundary.

**Fix:** Use the encoded length-field width—8 bytes for the 64-byte TLS hashes and 16 bytes for the 128-byte hashes—when detecting whether final padding needs another block. Add boundary cases for real message lengths 55/56 and 111/112.

**Suggestion:**
```suggestion
padSz = blockSz >> 3;
```

src/wp_hmac.c
```diff
@@ -51,6 +51,11 @@ typedef struct wp_HmacCtx {
     unsigned char* key;
     /** Length of private key in bytes. */
     size_t keyLen;
+
+    /** Length of the padded TLS record including MAC and padding. */
```
aidangarske, line 55: **Preserve TLS state when duplicating HMAC contexts**

The PR adds `tlsDataSize` and `tlsDummyBlocks`, but `wp_hmac_dup()` manually copies only the pre-existing fields. Because the destination is zero-allocated, `EVP_MAC_CTX_dup()` or `EVP_MD_CTX_copy_ex()` after setting `TLS_DATA_SIZE` loses the configured record size and computed dummy count. The duplicate therefore finalizes without the new TLS equalization behavior, violating the provider's `dupctx` state-copy contract. The existing duplication tests do not set this new parameter.

**Fix:** Copy both TLS fields in `wp_hmac_dup()` and add a regression that duplicates the context after setting `TLS_DATA_SIZE`, including a duplicate taken after the second update.

test/test_hmac.c
```diff
@@ -768,5 +769,107 @@ int test_hmac_size_query(void *data)
...
+    return ret;
+}
+
+/* Setting TLS_DATA_SIZE must not change the record MAC. */
+int test_hmac_tls_data_size(void *data)
+{
+    int ret = 0;
+    unsigned char key[32];
+    unsigned char header[13];
+    unsigned char msg[100];
```
aidangarske, line 814: **Exercise the TLS block-count behavior in the regression test**

The test only compares the resulting MAC. Dummy hashing occurs after `wc_HmacFinal()`, so its count cannot affect those bytes; the test still passes if the block calculator or dummy loop is removed entirely. It also advertises a 148-byte TLS record while the second update is backed by only `msg[100]`, despite the `TLS_DATA_SIZE` contract requiring the pointer to cover the entire record, including MAC and padding.

**Fix:** Back the TLS update with a complete, block-aligned record buffer and directly validate compression/update counts through instrumentation or a focused unit test of the block-count helper. Cover both sides of each hash-padding boundary.

src/wp_aes_stream.c
```diff
@@ -308,6 +311,13 @@ static int wp_aes_stream_init(wp_AesStreamCtx *ctx, const unsigned char *key,
     if (ok && (iv != NULL) && (!wp_aes_init_iv(ctx, iv, ivLen))) {
         ok = 0;
     }
+    if (ok && (iv == NULL) && ctx->ivSet &&
```
aidangarske, line 314: **Cover the new CFB no-IV reinitialization branch**

The new original-IV restoration condition changes AES-CFB as well as AES-CTS, but the added split-initialization test fetches only `AES-128-CBC-CTS`. Existing CFB stream tests always provide an explicit IV on initialization, so the new CFB branch is not exercised.

**Fix:** Add an AES-CFB encrypt/decrypt regression that processes data, reinitializes the same context with a NULL IV, and verifies that output matches OpenSSL and the first run.

aidangarske (changes requested):
## Skoll Code Review

**Scan type:** `review`

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [High] **Use the length-field width when counting final hash blocks** — `src/wp_hmac.c:313-321`
- [High] **Preserve TLS state when duplicating HMAC contexts** — `src/wp_hmac.c:55-58`
- [Medium] **Exercise the TLS block-count behavior in the regression test** — `test/test_hmac.c:814-863`
- [Medium] **Cover the new CFB no-IV reinitialization branch** — `src/wp_aes_stream.c:314-320`

src/wp_hmac.c
```diff
@@ -258,6 +265,64 @@ static int wp_hmac_init(wp_HmacCtx* macCtx, const unsigned char* key,
...
+{
+    int blockSizeRet = wc_HashGetBlockSize(hashType);
+    word32 blockSz;
+    word32 blockMask;
+    word32 padSz;
+    word32 realSz;
+    word32 maxSz;
+    int blockBits = 0;
+
+    if ((blockSizeRet <= 0) || (tlsDataSize < dataLen + macSize + 1)) {
```
aidangarske: **Invalid TLS padding takes a distinguishable HMAC timing path** · _Security_

BEFORE, this helper and branch did not exist. AFTER, it returns immediately when `tlsDataSize < dataLen + macSize + 1`. The provider's CBC unpadding path leaves `dataLen == tlsDataSize - macSize` when padding is invalid, making the condition true, while every valid record removes at least one padding byte and executes the block-count loop. OpenSSL defines this parameter specifically for received MAC-then-encrypt CBC records and supplies the post-unpadding length ([EVP_MAC documentation](https://docs.openssl.org/3.5/man3/EVP_MAC/), [libssl caller](https://github.com/openssl/openssl/blob/master/ssl/record/methods/tls1_meth.c)). Consequently, attacker-chosen ciphertext controls whether the newly added helper takes its early return based on padding validity. Both cases subsequently fail MAC verification, so the runtime difference can become a TLS CBC padding oracle and undermines the intended constant-time mitigation.

**Fix:** Allow the invalid-padding equality case through the same block-count computation as valid records, while rejecting only values outside the TLS contract. Add a regression covering valid minimum padding and invalid padding at the same padded record length.

**Suggestion:**
```suggestion
if ((blockSizeRet <= 0) || (tlsDataSize <= macSize) ||
        (dataLen > tlsDataSize - macSize)) {
    return 1;
}
```

test/test_hmac.c
```diff
@@ -768,5 +771,166 @@ int test_hmac_size_query(void *data)
...
+int test_hmac_tls_dummy_blocks(void *data)
+{
+    static const struct {
+        enum wc_HashType type;
+        size_t macSize;
+        size_t blockSz;
+        size_t lenSz;
+    } hashes[] = {
+        { WC_HASH_TYPE_SHA256, 32, 64,  8  },
+        { WC_HASH_TYPE_SHA384, 48, 128, 16 },
```
aidangarske, line 898: **SHA-384 dummy-block test runs when SHA-384 is disabled** · _Regression_

BEFORE, SHA-384 HMAC tests in this file were guarded by `WP_HAVE_SHA384`. AFTER, the new test unconditionally includes `WC_HASH_TYPE_SHA384`, although `WP_HAVE_HMAC` and `WP_HAVE_SHA384` are independent configuration gates. In an HMAC-enabled build without SHA-384, `wc_HashGetBlockSize()` reports the algorithm as unavailable, the helper always returns `1`, and this test fails when its expected SHA-384 block count changes. Some supported wolfCrypt versions may instead fail compilation if the enum is gated out.

**Fix:** Guard the SHA-384 test vector with `WP_HAVE_SHA384` and verify an HMAC-enabled, SHA-384-disabled configuration.

**Suggestion:**
```suggestion
#ifdef WP_HAVE_SHA384
        { WC_HASH_TYPE_SHA384, 48, 128, 16 },
#endif
```

aidangarske (dismissed):
## Skoll Code Review

**Scan type:** `review-security`

### Posted findings
- [High] **Invalid TLS padding takes a distinguishable HMAC timing path** — `src/wp_hmac.c:308-310`
- [Low] **SHA-384 dummy-block test runs when SHA-384 is disabled** — `test/test_hmac.c:897`

aidangarske:
Jenkins retest this please

gasbytes:
Jenkins retest this please

src/wp_hmac.c
```diff
@@ -323,6 +393,24 @@ static int wp_hmac_final(wp_HmacCtx* macCtx, unsigned char* out, size_t* outl,
...
+        int blockSz = wc_HashGetBlockSize(macCtx->type);
+        int i;
+
+        if (blockSz <= 0) {
+            ok = 0;
+        }
+        XMEMSET(dummy, 0, sizeof(dummy));
+        for (i = 0; ok && (i < macCtx->tlsDummyBlocks); i++) {
+            rc = wc_HmacUpdate(&macCtx->hmac, dummy, (word32)blockSz);
+            if (rc != 0) {
```
padelsbach, line 455: This does a hmac operation after `wc_HmacFinal` which resets the object. We might get bogus or non-deterministic results. Can you check?

src/wp_hmac.c
```diff
@@ -274,6 +339,11 @@ static int wp_hmac_update(wp_HmacCtx* macCtx, const unsigned char* data,
 
     WOLFPROV_ENTER(WP_LOG_COMP_MAC, "wp_hmac_update");
 
+    if (macCtx->tlsDataSize > 0) {
```
padelsbach: Does this work on chunk'ed data, or combined header+data? Seems like the block count is reestablished on each call

src/wp_hmac.c
```diff
@@ -258,6 +265,64 @@ static int wp_hmac_init(wp_HmacCtx* macCtx, const unsigned char* key,
...
+ * @param [in] len        Length of message in bytes.
+ * @param [in] blockBits  Log base 2 of the hash block size.
+ * @param [in] blockMask  Hash block size minus one.
+ * @param [in] padSz      Bytes of padding the hash appends to the message.
+ * @return  Number of blocks processed.
+ */
+static int wp_hmac_blocks(word32 len, int blockBits, word32 blockMask,
+    word32 padSz)
+{
+    return (int)(len >> blockBits) +
```
padelsbach, line 324: Does the shift arithmetic work for HMAC-SHA3?

padelsbach (changes requested):

src/wp_hmac.c
gasbytes: the mac itself is safe since it's already written to the output buffer before the dummy hashing starts, but it's true that it leaves the context "dirty" instead of clean, so I'll move the dummy blocks onto a scratch copy of the state.

src/wp_hmac.c
gasbytes: thanks, i've added a function that checks for that based on hash type, so finalize now fails rather than equalizing with the wrong count

test/test_cipher.c
```diff
@@ -1459,6 +1525,157 @@ int test_aes256_cts(void *data)
...
+
+/* CTS must use the cached IV when the key and IV are set by separate init
+ * calls, in either order. */
+int test_aes128_cts_split_init(void *data)
+{
+    int err;
+    int keyFirst;
+    unsigned char key[16];
+    unsigned char iv[16];
+    unsigned char msg[17];
```
padelsbach, line 1590: can we add a combined test case of both one block + split init? I think this is needed to complete the test coverage.

src/wp_hmac.c
padelsbach: can you add a test case for this? A simple check could be that a second init from the same context should yield the same result. Eg:

```
EVP_MAC_init(ctx, key, sizeof(key), dp);
// Mac update + final
EVP_MAC_init(ctx, NULL, 0, NULL);
// Mac update + final
// memcmp outputs, should match
```

test/test_cipher.c
gasbytes: thanks! added in the latest squashed commit.

padelsbach (approved):

PR #477, opened 2026-08-26
dh: allow a zero-size export buffer

yosuke-wolfssl:
### Problem
`wp_dh_export()` computes a single allocation covering both the group data and the key pair, then requires `OPENSSL_secure_malloc()` to return non-NULL before it exports anything. For a parameter-only named-group key, both contributions are zero: `wp_dh_export_group_alloc_size()` returns 0 because a named group is exported as a string pointer with no backing buffer, and `wp_dh_export_keypair_alloc_size()` returns 0 when there are no public or private bytes. `OPENSSL_secure_malloc(0)` returns NULL, so a key that needed no buffer at all fails to export.

Reachable through `EVP_PKEY_todata()` and through any cross-provider export of DH domain parameters. Closes f-11557.

### Fix (`src/wp_dh_kmgmt.c`)
Allocate only when the computed size is non-zero:

```c
        /* Named group parameters need no buffer - zero size is valid. */
        if (sz > 0) {
            data = OPENSSL_secure_malloc(sz);
            if (data == NULL) {
                ok = 0;
            }
        }
```

`data` stays NULL on that path and is never dereferenced: `wp_dh_export_group()` touches it only in its explicit-parameters branch, `wp_dh_export_keypair()` is guarded by `pubSz > 0` / `privSz > 0`, and `OPENSSL_secure_clear_free()` returns early on a null pointer.

### Tests
`test_dh_export_named_group_params` imports an `ffdhe2048` parameter-only key via `EVP_PKEY_fromdata()`, exports it with `EVP_PKEY_todata()`, and requires a group name in the result — so it fails on a silently empty export as well as on the error return.

### Verification
- Build clean under `-Werror`; no new compiler warnings.
- Full unit suite: 200/200 pass. DH cases: 15/15.
- Negative control: with the guard reverted the new test fails at `Export of parameter-only named group key failed`; it passes with the guard in place.

### DIFF
--- a/src/wp_dh_kmgmt.c
+++ b/src/wp_dh_kmgmt.c
@@ -1498,6 +1498,11 @@ static int wp_dh_export(wp_Dh *dh, int selection, OSSL_CALLBACK *paramCb,
     if (ok && (dh == NULL)) {
         ok = 0;
     }
+    /* Nothing to export when the key has no domain parameters. */
+    if (ok && ((selection & OSSL_KEYMGMT_SELECT_ALL_PARAMETERS) != 0) &&
+            (!wp_dh_has(dh, OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS))) {
+        ok = 0;
+    }
     if (ok) {
         XMEMSET(params, 0, sizeof(params));
 
@@ -1506,9 +1511,12 @@ static int wp_dh_export(wp_Dh *dh, int selection, OSSL_CALLBACK *paramCb,
             sz += wp_dh_export_group_alloc_size(dh);
         }
 
-        data = OPENSSL_secure_malloc(sz);
-        if (data == NULL) {
-            ok = 0;
+        /* Named group parameters need no buffer - zero size is valid. */
+        if (sz > 0) {
+            data = OPENSSL_secure_malloc(sz);
+            if (data == NULL) {
+                ok = 0;
+            }
         }
     }
     if (ok && ((selection & OSSL_KEYMGMT_SELECT_ALL_PARAMETERS) != 0)) {
--- a/test/test_dh.c
+++ b/test/test_dh.c
@@ -1735,4 +1735,67 @@ int test_dh_pgen_min_bits(void *data)
     return err;
 }
 
+/*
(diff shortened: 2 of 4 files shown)


src/wp_dh_kmgmt.c
```diff
@@ -1506,9 +1506,12 @@ static int wp_dh_export(wp_Dh *dh, int selection, OSSL_CALLBACK *paramCb,
             sz += wp_dh_export_group_alloc_size(dh);
         }
 
-        data = OPENSSL_secure_malloc(sz);
-        if (data == NULL) {
-            ok = 0;
+        /* Named group parameters need no buffer - zero size is valid. */
+        if (sz > 0) {
```
yosuke-wolfssl, line 1515: Thanks — the structural point is right, though the specific fault doesn't occur.

`wp_param_set_mp()` sets `data_size = mp_unsigned_bin_size(&dh->key.p)`, which is `0` for a zero `p`, then calls `mp_to_unsigned_bin(mp, NULL)`. wolfSSL's `sp_to_unsigned_bin_len()` validates `out == NULL` and returns `MP_VAL` without dereferencing, so the export already failed cleanly. Verified at runtime; ASan + UBSan clean.

Still worth fixing: before this PR, `data == NULL` always implied `ok == 0`, so `wp_dh_export_group()` could never run without a buffer. The `sz > 0` guard broke that. `wp_dh_export()` now rejects the key up front instead:

```c
    /* Nothing to export when the key has no domain parameters. */
    if (ok && ((selection & OSSL_KEYMGMT_SELECT_ALL_PARAMETERS) != 0) &&
            (!wp_dh_has(dh, OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS))) {
        ok = 0;
    }
```

`wp_dh_has()` already computes `(dh->id != 0) || (!mp_iszero(&dh->key.p))`, so the check can't drift from it.

yosuke-wolfssl, line 1515: Same observation as the sibling comment, addressed by the same change — `wp_dh_export()` now rejects a key with no domain parameters via `wp_dh_has()`, so `data == NULL` can no longer reach the explicit-parameters branch.

PR #478, opened 2026-08-27
Fix EdDSA lazy public-key derivation race on shared private-only keys

sameehj:
## Summary

A private-only EdDSA key (a seed-only PKCS#8 Ed25519 or Ed448 key) has no
public half after import. wolfProvider derives the public half later, on
first use. The derivation writes into the shared wolfSSL key object without
holding the key mutex. When two or more threads first use the same
`EVP_PKEY` at the same time, they sign or export with a partly written
public key. This produces failed or invalid signatures.

Reported by: Fenrir finding 11559.

## Root cause

`wc_ed25519_make_public` / `wc_ed448_make_public` set `pubKeySet` when they
write to the output buffer. They do not always fill `key->p`. Only
`wc_ed*_import_public` stores the value into `key->p`. The old code derived
the public half in three places without the mutex:

- `wp_ed25519_export_public` / `wp_ed448_export_public`
- `wp_ed25519_digest_sign` / `wp_ed448_digest_sign` (derived before the lock;
  only `wc_ed*_sign_msg` was locked)
- `wp_Ed25519PublicKeyToDer` / `wp_Ed448PublicKeyToDer`

A concurrent first use could read `key->p` while another thread wrote it.

## Fix

Derive the public half in one place, under the key mutex.

- Add a `derivePub` callback to `wp_EcxData` (set for Ed25519 and Ed448,
  `NULL` for X25519/X448).
- Add `wp_ecx_ensure_pub()`. It takes the key mutex, then calls the derive
  helper. The helper calls `make_public` into a local buffer, then
  `import_public`, which is the only call that sets `key->p` and `pubKeySet`.
- Remove the unlocked derivation from the export and DER helpers. They are
  now plain export/encode.
- Call `wp_ecx_ensure_pub()` from every first-use site: the sign paths,
  `wp_ecx_get_params_enc_pub_key`, `wp_ecx_match_pub_key`,
  `wp_ecx_export_keypair`, `wp_ecx_dup`, and the SPKI branch of
  `wp_ecx_encode`.

The SPKI-only guard in `wp_ecx_encode` keeps the public key out of a private
key encoding. In `wp_ecx_dup`, the derive runs before the key copy, so the
copy also gets a happens-before edge against a concurrent first use.

## Scope: other algorithms

I reviewed the other key types for the same pattern.

- ECC also derives a public key from a private key, but it does so at
  import/decode time, while the object has one owner. No race.
- ML-DSA does not derive on first use. It reports `hasPub = 0` for a
  private-only key and fails cleanly.
- RSA, DH, ECDSA, and ML-KEM have no first-use public derivation.

EdDSA was the only affected type.

## Test

`test_ecx_shared_key_first_use` (test 195) loads one seed-only key, then uses
it from 4 threads at the same time. Half the threads sign, half export. The
signers verify against a separate public-only key, so a bad public half
cannot hide a bad signature. The workload is fixed and small, so any failure
is a defect, not timing.

- Without the fix: the test fails on every run (10/10 local runs).
- With the fix: the test passes, and the full unit suite passes.

ThreadSanitizer (the `tsan` job) is the reliable detector for the data race
itself. The test is the deterministic correctness guard.

## Severity

Medium. The window is first use of a shared private-only key. The impact is
failed or invalid signatures, not key disclosure or forgery. A TLS load path
that warms the public key via `X509_check_private_key` is not affected. A
sign-only service with a raw private import and a thread pool is the
realistic case.

## Commits

- `test:` adds the failing test.
- `fix:` adds the fix.

The first commit fails on its own by design (red), the second makes it pass
(green). Please merge as a unit; do not run per-commit CI or bisect across
the pair.

### DIFF
--- a/src/wp_ecx_kmgmt.c
+++ b/src/wp_ecx_kmgmt.c
@@ -85,6 +85,8 @@ typedef int (*WP_ECX_IMPORT_PRIV)(const byte* priv, word32 privLen, void* key,
     int endian);
 /** Type for function that exports a private key from a wolfSSL key. */
 typedef int (*WP_ECX_EXPORT_PRIV)(void* key, const byte* out, word32* outLen);
+/** Type for function that derives the public key from the private key. */
+typedef int (*WP_ECX_DERIVE_PUB)(void* key);
 /** Type for a wolfSSL function that checks a public key. */
 typedef int (*WP_ECX_CHECK_PUB)(const byte* pub, word32 pubLen, int endian);
 /** Type for a wolfSSL function that checks a key. */
@@ -117,6 +119,8 @@ typedef struct wp_EcxData {
     WP_ECX_IMPORT_PRIV importPriv;
     /** Export private key from wolfSSL key object. */
     WP_ECX_EXPORT_PRIV exportPriv;
+    /** Derive the public key from the private key. NULL when not needed. */
+    WP_ECX_DERIVE_PUB  derivePub;
     /** wolfSSL check of public key value. */
     WP_ECX_CHECK_PUB   checkPub;
     /** wolfSSL check of public/private key. */
@@ -259,6 +263,48 @@ wolfSSL_Mutex* wp_ecx_get_mutex(wp_Ecx* ecx)
 }
 
 /**
+ * Make sure the public key is available in the wolfSSL key object.
+ *
+ * Ed25519 and Ed448 keys imported as a private key only have no public half
+ * until it is derived. Deriving writes into the wolfSSL key object, so the
+ * key mutex is held: threads that first use one shared key at the same time
+ * would otherwise sign or export with a partly written public key.
+ *
+ * Must not be called with the key mutex already held.
+ *
+ * @param [in, out] ecx  ECX key object.
+ * @return  1 on success.
+ * @return  0 on failure.
+ */
+int wp_ecx_ensure_pub(wp_Ecx* ecx)
+{
+    int ok = 1;
+
+    if ((ecx != NULL) && (ecx->data->derivePub != NULL)) {
+        int rc;
+
+    #ifndef WP_SINGLE_THREADED
+        if (wp_lock(&ecx->mutex) != 1) {
+            ok = 0;
+        }
+        if (ok) {
+    #endif
+            rc = (*ecx->data->derivePub)((void*)&ecx->key);
+            if (rc != 0) {
+                WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "derivePub",
+                    rc);
+                ok = 0;
+            }
+    #ifndef WP_SINGLE_THREADED
+            wp_unlock(&ecx->mutex);
+        }
+    #endif
+    }
+
+    return ok;
+}
+
+/**
  * Create a new ECX key object. Base function.
  *
  * @param [in] provCtx   Provider context.
@@ -363,6 +409,13 @@ static wp_Ecx* wp_ecx_dup(const wp_Ecx* src, int selection)
     if (wolfssl_prov_is_running()) {
         /* Create a new ecx object. */
         dst = wp_ecx_new(src->provCtx, src->data);
+    }
+    /* Cache the public half before the copy below: deriving it later writes
+     * into the source key object that is being read here. */
+    if ((dst != NULL) && src->hasPub &&
+            (!wp_ecx_ensure_pub((wp_Ecx*)src))) {
+        wp_ecx_free(dst);
+        dst = NULL;
     }
     if (dst != NULL) {
         dst->includePublic = src->includePublic;
@@ -552,9 +605,18 @@ static int wp_ecx_get_params_enc_pub_key(wp_Ecx* ecx, OSSL_PARAM params[],
         if (p->data == NULL) {
             outLen = ecx->data->len;
         }
+        else if (!wp_ecx_ensure_pub(ecx)) {
+            ok = 0;
+        }
+        else if (wp_lock(wp_ecx_get_mutex(ecx)) != 1) {
+            ok = 0;
+        }
         else {
+            /* Signing mutates the key (persistent SHA), so hold the key mutex
+             * while reading it. */
             int rc = (*ecx->data->exportPub)((void*)&ecx->key, p->data,
                 &outLen, ECX_LITTLE_ENDIAN);
+            wp_unlock(wp_ecx_get_mutex(ecx));
             if (rc != 0) {
                 WOLFPROV_MSG_DEBUG_RETCODE(WP_LOG_LEVEL_DEBUG, "exportPub", rc);
                 ok = 0;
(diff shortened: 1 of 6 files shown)


test/test_ecx.c
```diff
@@ -1174,6 +1178,215 @@ int test_ecx_dup(void *data)
...
+} wp_ecx_shared_args;
+
+/*
+ * Sign with the shared key and verify against a separate public-only key, so
+ * a bad public half on the shared key cannot hide a bad signature.
+ */
+static int wp_ecx_shared_sign(wp_ecx_shared_args* w)
+{
+    int err;
+    unsigned char sig[ED448_SIG_SIZE];
```
sameehj: Fixed in af3b531. The shared helpers now size sig[] and pub[] from WP_ECX_SHARED_SIG_SIZE / WP_ECX_SHARED_PUB_SIZE, which pick ED25519 sizes when WP_HAVE_ED448 is off, ED448 sizes when only Ed448 is built, and MAX of both when both are built — the same pattern as the readback buffers earlier in this file.

These two comments are the same compile break; the Memory management label does not apply. I did not add a CI cell for Ed25519-only wolfSSL: a unit test cannot catch a compile failure that current --enable-all-crypto jobs never hit. Verified locally that the Ed25519-only branch expands to ED25519_SIG_SIZE / ED25519_PUB_KEY_SIZE.

sameehj: Fixed in af3b531. The shared helpers now size sig[] and pub[] from WP_ECX_SHARED_SIG_SIZE / WP_ECX_SHARED_PUB_SIZE, which pick ED25519 sizes when WP_HAVE_ED448 is off, ED448 sizes when only Ed448 is built, and MAX of both when both are built — the same pattern as the readback buffers earlier in this file.

These two comments are the same compile break; the Memory management label does not apply. I did not add a CI cell for Ed25519-only wolfSSL: a unit test cannot catch a compile failure that current --enable-all-crypto jobs never hit. Verified locally that the Ed25519-only branch expands to ED25519_SIG_SIZE / ED25519_PUB_KEY_SIZE.

test/test_ecx.c
```diff
@@ -1174,6 +1178,227 @@ int test_ecx_dup(void *data)
...
+            if (pthread_create(&tids[i], NULL, wp_ecx_shared_key_thread,
+                    &workers[i]) != 0) {
+                PRINT_ERR_MSG("Failed to create thread %d", i);
+                err = 1;
+                break;
+            }
+            created++;
+        }
+
+        /* Release every worker at once so they hit the first use together. */
```
aidangarske: **Start gate does not wait for every worker**

The new test broadcasts immediately after creating the threads, without confirming that every worker has reached the condition-variable wait. A delayed worker can observe `start == 1` and proceed only after another worker has completed the first-use derivation. Consequently, the test is scheduler-dependent and can pass against the unfixed implementation despite claiming to release all workers together.

**Fix:** Track a ready-worker count under `startLock`. Have each worker increment it before waiting, and have the parent wait until all created workers are ready before setting `start` and broadcasting.

test/test_ecx.c
```diff
@@ -1174,6 +1178,227 @@ int test_ecx_dup(void *data)
...
+    int err;
+    unsigned char sig[WP_ECX_SHARED_SIG_SIZE];
+    size_t sigLen = w->sigLen;
+    unsigned char msg[] = WP_ECX_SHARED_MSG;
+    EVP_PKEY* pubKey = NULL;
+
+    err = test_digest_sign(w->key, wpLibCtx, msg, sizeof(msg) - 1, NULL, NULL,
+        sig, &sigLen, 0, 0);
+    if (err == 0) {
+        pubKey = EVP_PKEY_new_raw_public_key_ex(osslLibCtx, w->algName, NULL,
```
aidangarske, line 1244: **Concurrent wolfProvider verification is not exercised**

This PR adds mutex protection to `wp_ed25519_digest_verify` and `wp_ed448_digest_verify`, but the new threaded test creates its verification key in `osslLibCtx` and verifies through the OpenSSL default provider. The changed wolfProvider verification callbacks therefore never run concurrently, so an omitted or incorrect verify lock would remain invisible to both this test and TSan.

**Fix:** Add a contention case using a shared wolfProvider-backed public key and a precomputed valid signature, with multiple workers verifying concurrently.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [Medium] **Start gate does not wait for every worker** — `test/test_ecx.c:1341-1345`
- [Medium] **Concurrent wolfProvider verification is not exercised** — `test/test_ecx.c:1231-1237`

test/test_ecx.c
sameehj: Addressed, thanks.

test/test_ecx.c
sameehj: Thank you, addressed.

src/wp_ecx_kmgmt.c
padelsbach, line 429: We probably want to lock the mutex while doing this memcpy and `exportPub`. This is a pre-existing issue

src/wp_ecx_kmgmt.c
```diff
@@ -258,6 +262,48 @@ wolfSSL_Mutex* wp_ecx_get_mutex(wp_Ecx* ecx)
...
+ * @return  0 on failure.
+ */
+int wp_ecx_ensure_pub(wp_Ecx* ecx)
+{
+    int ok = 1;
+
+    if ((ecx != NULL) && (ecx->data->derivePub != NULL)) {
+        int rc;
+
+    #ifndef WP_SINGLE_THREADED
```
padelsbach: Nit: the check of `WP_SINGLE_THREADED` is built into `wp_lock`. Other callers in WP of `wp_lock` don't do this check

src/wp_ecx_kmgmt.c
```diff
@@ -552,9 +605,18 @@ static int wp_ecx_get_params_enc_pub_key(wp_Ecx* ecx, OSSL_PARAM params[],
...
             outLen = ecx->data->len;
         }
+        else if (!wp_ecx_ensure_pub(ecx)) {
+            ok = 0;
+        }
+        else if (wp_lock(wp_ecx_get_mutex(ecx)) != 1) {
+            ok = 0;
+        }
         else {
+            /* Signing mutates the key (persistent SHA), so hold the key mutex
```
padelsbach: What is the "persistent SHA" referring to in this case (and a couple similar comments)? Not sure a comment is even needed here

src/wp_ecx_kmgmt.c
```diff
@@ -1650,20 +1749,33 @@ static int wp_ed25519_import_public(const byte* in, word32 inLen,
...
+ * @return  -ve on failure.
+ */
+static int wp_ed25519_derive_public(ed25519_key* key)
+{
+    int ret = 0;
+    byte pub[ED25519_PUB_KEY_SIZE];
+
     if (!key->pubKeySet) {
-        ret = wc_ed25519_make_public(key, (byte*)out, *outLen);
+        ret = wc_ed25519_make_public(key, pub, sizeof(pub));
```
padelsbach, line 1782: AI says it may be more efficient to use `wc_ed25519_import_public_ex` with the `trusted` param set. Can you investigate?

padelsbach (commented):
Looks good, couple small questions/concerns

src/wp_ecx_kmgmt.c
sameehj: Done, mutex added.

src/wp_ecx_kmgmt.c
sameehj: Removed, using wp_lock.

src/wp_ecx_kmgmt.c
sameehj: Clarified, referencing macro.

src/wp_ecx_kmgmt.c
sameehj: Done, trusted=1.

padelsbach (approved):
Looks good, thanks

PR #482, opened 2026-08-28
Stop the one-shot AES-CCM length pass from finalizing the operation

aidangarske:
after LMS merged the new CAVP test caught new regression on master

### DIFF
--- a/src/wp_aes_aead.c
+++ b/src/wp_aes_aead.c
@@ -2313,12 +2313,13 @@ static int wp_aesccm_cipher(wp_AeadCtx *ctx, unsigned char *out,
     }
     if (ok) {
         *outLen = 0;
-        /* Single-shot CCM: a both-NULL pass before the encrypt (tagAvail clear)
-         * is only the one-shot total-length declaration and must not finalise,
-         * or the payload encrypt that follows is rejected. After the encrypt it
-         * is a trailing finalise and falls through so IV_STATE_FINISHED is set
-         * against nonce reuse. */
-        if ((in == NULL) && (out == NULL) && (!ctx->tagAvail)) {
+        /* One-shot CCM: the total-length declaration and the trailing finalise
+         * both arrive as a NULL/NULL pass. ivSet (set by the payload encrypt,
+         * and unlike tagAvail not cleared by GET_TAG) separates them: clear is
+         * the length declaration and must not finalise, or the payload that
+         * follows is rejected; set falls through to finalise so
+         * IV_STATE_FINISHED is set against nonce reuse. */
+        if ((in == NULL) && (out == NULL) && (!ctx->ivSet)) {
             *outLen = inLen;
         }
         else if (in != NULL) {
--- a/src/wp_aes_aead.c
+++ b/src/wp_aes_aead.c
@@ -2313,7 +2313,15 @@ static int wp_aesccm_cipher(wp_AeadCtx *ctx, unsigned char *out,
     }
     if (ok) {
         *outLen = 0;
-        if (in != NULL) {
+        /* Single-shot CCM: a both-NULL pass before the encrypt (tagAvail clear)
+         * is only the one-shot total-length declaration and must not finalise,
+         * or the payload encrypt that follows is rejected. After the encrypt it
+         * is a trailing finalise and falls through so IV_STATE_FINISHED is set
+         * against nonce reuse. */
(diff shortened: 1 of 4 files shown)


src/wp_aes_aead.c
```diff
@@ -2313,7 +2313,14 @@ static int wp_aesccm_cipher(wp_AeadCtx *ctx, unsigned char *out,
     }
     if (ok) {
         *outLen = 0;
-        if (in != NULL) {
+        /* wolfCrypt CCM is single-shot: a both-NULL pass (the one-shot
+         * total-length declaration, or a trailing no-op finalise after the
+         * encrypt) has nothing to buffer or finalise, so report success rather
+         * than routing it into the tagAvail-guarded update or the finalise. */
+        if ((in == NULL) && (out == NULL)) {
```
padelsbach: AI says that the old behavior would set `ctx->ivState = IV_STATE_FINISHED;` when `NULL, NULL` was passed in, and this behavior is lost with the change. Seems like we should keep that to avoid IV reuse

test/test_aestag.c
```diff
@@ -2466,6 +2466,128 @@ int test_aes_ccm_ctx_reuse(void *data)
...
+    err = cipher == NULL;
+
+    /* Encrypt through the one-shot EVP_Cipher() API a caller such as libacvp
+     * uses: a NULL/NULL total-length pass, an AAD pass, then the payload. The
+     * length pass must not be mistaken for a finalise or the payload fails. */
+    if (err == 0) {
+        err = (ctx = EVP_CIPHER_CTX_new()) == NULL;
+    }
+    if (err == 0) {
+        err = EVP_CipherInit(ctx, cipher, NULL, NULL, 1) != 1;
```
padelsbach, line 2499: Can we test a one-shot decrypt too?

test/test_aestag.c
```diff
@@ -2466,6 +2466,128 @@ int test_aes_ccm_ctx_reuse(void *data)
...
+    unsigned char iv[13];
+    unsigned char aad[] = "additional data";
+    unsigned char pt[] = "CCM one-shot plaintext";
+    int ptLen = (int)(sizeof(pt) - 1);
+    unsigned char ct[64];
+    unsigned char tag[16];
+    unsigned char dec[64];
+    int decLen = 0, fLen = 0;
+
+    memset(key, 0xAA, keyLen);
```
padelsbach: Nit: can we use `sizeof(key)` instead of `keyLen` for consistency and memory safety

src/wp_aes_aead.c
aidangarske: makes sense kept it

test/test_aestag.c
aidangarske: yes

aidangarske:
Jenkins retest this please 
(403 error)

src/wp_aes_aead.c
```diff
@@ -2313,7 +2313,15 @@ static int wp_aesccm_cipher(wp_AeadCtx *ctx, unsigned char *out,
     }
     if (ok) {
         *outLen = 0;
-        if (in != NULL) {
+        /* Single-shot CCM: a both-NULL pass before the encrypt (tagAvail clear)
+         * is only the one-shot total-length declaration and must not finalise,
+         * or the payload encrypt that follows is rejected. After the encrypt it
+         * is a trailing finalise and falls through so IV_STATE_FINISHED is set
+         * against nonce reuse. */
+        if ((in == NULL) && (out == NULL) && (!ctx->tagAvail)) {
```
padelsbach: `tagAvail` is cleared by `EVP_CTRL_AEAD_GET_TAG`, so doing this sequence still looks wrong:

1. encrypt: `tagAvail` becomes 1
2. `GET_TAG`, `tagAvail` back to 0
3. `in = out = NULL`
4. encrypt again: returns 22

padelsbach (changes requested):

src/wp_aes_aead.c
aidangarske: wow good catch thank you it now discrims on the ivSet now

padelsbach (approved):

aidangarske:
Jenkins retest this please

padelsbach:
jenkins retest this please

padelsbach (approved):

